PR #21, opened 2017-08-15
Access control

Recmo:


contracts/Curve.sol
```diff
@@ -21,74 +20,41 @@ contract Curve is Ownable {
...
     {
-        totalEuroUlps = 0;
         NEUMARK_CONTROLLER = neumarkController;
+        totalEuroUlps = 0;
     }
 
     // issues to msg.sender, further neumarks distribution happens there
     function issue(uint256 euroUlps)
-        //onlyOwner()
+        only(ROLE_NEUMARK_ISSUER)
```
Recmo: Yes, sorry. It's not pushed yet.

contracts/AccessControl/RoleBasedAccessControl.sol
```diff
@@ -0,0 +1,147 @@
...
+    // Constants
+    ////////////////
+
+    IAccessControlled public constant GLOBAL = IAccessControlled(0x0);
+
+    ////////////////
+    // State
+    ////////////////
+
+    mapping (address => mapping(bytes32 => mapping(address => TriState))) access;
```
Recmo: Not too much more complicated. I added a query `getUsers(object, role)` that returns all users that have a permission set (either allow or deny).

contracts/AccessControl/RoleBasedAccessControl.sol
```diff
@@ -0,0 +1,147 @@
...
+        bytes32 role,
+        TriState access
+    )
+        public
+    {
+        return setUserRole(subject, role, GLOBAL, access);
+    }
+
+    // Assign a role to a user globally
+    function setUserRole(
```
Recmo: Correct. You can always turn these two transactions into one using a helper wallet contract, but it will get a bit tricky with the rights to do so. I can easily add a batch operation:

```
    function setUserRole(
        address[] subjects,
        bytes32[] roles,
        IAccessControlled[] objects,
        TriState[] newValues
    )
```

Then you can do all sorts of complex operations atomically.

PR #27, opened 2017-08-17
Access control

Recmo:


contracts/AccessControl/AccessControlled.sol
```diff
@@ -0,0 +1,42 @@
+pragma solidity ^0.4.11;
```
krzkaczor: mismatched pragma 0.4.11? `AccessControlExample.sol` has different one

PR #28, opened 2017-08-17
Penalty disbursal address

rudolfix:


contracts/ApproveAndCallCallback.sol
```diff
@@ -0,0 +1,5 @@
+pragma solidity 0.4.15;
+
+contract ApproveAndCallFallBack {
```
krzkaczor: make it an interface perhaps?

test/LockedAccount.js
```diff
@@ -6,10 +6,13 @@ import eventValue from "./helpers/eventValue";
 import * as chain from "./helpers/spawnContracts";
 import increaseTime, { setTimeTo } from "./helpers/increaseTime";
 import latestTime, { latestTimestamp } from "./helpers/latestTime";
+import EVMThrow from "./helpers/EVMThrow";
 
-const LockedAccount = artifacts.require("LockedAccount");
+const TestFeeDistributionPool = artifacts.require("TestFeeDistributionPool");
 
-contract("LockedAccount", accounts => {
+const should = require("chai").should();
```
krzkaczor: Just modify  1st line `import { expect, should } from "chai";`

test/LockedAccount.js
```diff
@@ -29,119 +32,242 @@ contract("LockedAccount", accounts => {
...
-      value: ticket / 2,
-      from: investor2,
-    });
+    // verify longstop date independently, value is convertable to int so do it
+    let unlockDate = timebase + 18 * 30 * chain.days;
+    if (parseInt(initialLockedBalance[2]) > 0) {
+      // earliest date is preserved for repeated investor address
+      unlockDate = parseInt(initialLockedBalance[2]);
+    }
+    assert.equal(parseInt(investorBalance[2]), unlockDate, "18 months in future");
```
krzkaczor: use expect

contracts/EtherToken.sol
```diff
@@ -14,6 +15,21 @@ contract EtherToken is StandardToken, TokenWithDeposit, Math {
     // disable default function
     function() { revert(); }
 
+    function approveAndCall(address _spender, uint256 _amount, bytes _extraData)
```
krzkaczor: shouldn't this be part of contract `ERC23`? This seems like quite generic implementation

contracts/LockedAccount.sol
```diff
@@ -95,57 +95,71 @@ contract LockedAccount is Ownable, TimeSource, ReturnsErrors, Math {
...
+        onlyStates(LockState.AcceptingUnlocks, LockState.ReleaseAll)
+        public
+        returns (Status)
+    {
+        return unlockFor(msg.sender);
+    }
+
     // this allows to unlock and allow neumarks to be burned in one transaction
     function receiveApproval(address from, uint256 _amount, address _token, bytes _data)
+        onlyStates(LockState.AcceptingUnlocks, LockState.ReleaseAll)
```
krzkaczor: what if its called not by a token contract?

contracts/EtherToken.sol
rudolfix: ERC23 is interface and should stay like that. I hope we can fix truffle

contracts/LockedAccount.sol
rudolfix: then you have this: `require(_token == address(neumarkToken));` check below. but a good catch! I have a test for this case as well

contracts/ERC23.sol
```diff
@@ -0,0 +1,7 @@
+pragma solidity ^0.4.11;
```
rudolfix: Is it still a problem? I think a separate PR with just this change should be better

contracts/ApproveAndCallCallback.sol
rudolfix: I copied this from remco's minimetoken.... We'll align it separaterly OK?

rudolfix: it was, then I bailed out on truffle problems

contracts/LockedAccount.sol
krzkaczor: i think we should more modifiers in a code, generally speaking

PR #30, opened 2017-08-18
Add tests coverage reports

krzkaczor:
Pay attention that it contains also changes from #29 branch (it's not yet merged).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,3 +47,5 @@ jspm_packages
 
 # Ignore jupyter / python notebooks
 .ipynb_checkpoints
+coverageEnv
+coverage.json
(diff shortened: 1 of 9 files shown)


Recmo:
While we are add it, should we add a linter for Solidity too? [Solium](https://www.npmjs.com/package/solium) seems to be the way to go.

There are a bunch more static analysis tools in the works (DappGuard/Oyente/SolFunctionProfiler), but these don't mature and seem ready for CI integration.

krzkaczor:
@Recmo for sure we have to do this! :)  I would just split it into separate pull requests as probably it will involve a lot of changes (i hope there is an autofix option) and discussion around certain rules.

PR #31, opened 2017-08-21
Whitelising tests

krzkaczor:


test/helpers/increaseTime.js
```diff
@@ -29,7 +29,7 @@ export default function increaseTime(duration) {
 }
 
 // sets time to given timestamp based on current block time
-export function setTimeTo(timestamp) {
+export async function setTimeTo(timestamp) {
```
Recmo: This function does not use `await`, so doesn't need `async`. This whole file could use a good rewrite, starting with a `toPromise` wrapper function for `web3` functions.

Recmo: Or maybe using the synchronous `send` functions like in `latestTimestamp()`.

test/helpers/latestTime.js
```diff
@@ -1,10 +1,36 @@
...
+export const DAY = 24 * 60 * 60;
+export const MONTH = HOUR * 24 * 31;
+
+// useful for spawning time sensitive contracts
+export function closeFutureDate() {
+  return latestTimestamp() + MONTH + MONTH;
+}
+
+// useful for spawning time sensitive contracts
+export function furterFutureDate() {
```
Recmo: Typo `furtHer`

test/helpers/latestTime.js
```diff
@@ -1,10 +1,36 @@
...
+  return latestTimestamp() + MONTH;
```
Recmo: Doesn't this make the further date closer than the closer date?

test/helpers/verification.js
```diff
@@ -0,0 +1,54 @@
...
+}
+
+let curve;
+
+export async function curveInEur(moneyInEur) {
+  if (!curve) {
+    curve = await deployCurve();
+  }
+
+  return curve.curve(moneyInEur);
```
Recmo: The function argument is in Euro Units of Least Precision (EuroULP), which is 10^-18 of a euro.

test/helpers/verification.js
```diff
@@ -0,0 +1,54 @@
...
+}
+
+export async function curveInEther(money, eurEtherRatio) {
+  if (!curve) {
+    curve = await deployCurve();
+  }
+
+  const moneyInEur = ethToEur(money, eurEtherRatio);
+
+  return curve.curve(moneyInEur);
```
Recmo: same here

test/helpers/verification.js
```diff
@@ -0,0 +1,54 @@
...
+}
+
+export function ethToEur(ether, eurEtherRatio = etherToWei(218.1192809)) {
```
Recmo: Make it 300. At least :p

test/helpers/increaseTime.js
krzkaczor: i added async there to make it clear that it's async ie. it returns a promise. I was a source of a bug in my previous tests

test/helpers/chaiWeb3.js
```diff
@@ -0,0 +1,26 @@
...
+    // I would love to hear ideas for better implementation
+
+    const web3ArrayAccessor = this._obj;
+
+    try {
+      await web3ArrayAccessor(size - 1);
+    } catch (e) {
+      this.assert(
+        false,
+        `expected web3 array to be size of ${size} but it looks like it's smaller` // i think it's impossible to get an array name in this point
```
Recmo: > impossible

```
    const req = web3ArrayAccessor.request(0);
    const bytes4 = req.params[0].data.slice(10);
```

The bytes4 contains a hash of the array name, you can now iterate through all contracts and find the matching one :D

test/helpers/increaseTime.js
Recmo: Yeah, my main problem is the inconsistency with `increaseTime` before it. That one should also have an `async` flag then.

Or, alternatively, replace `sendAsync` with `send` and make both function synchronous. This is how it is done elswhere (specifically in latestTimestamp). Again, for consistency we should do the same everywhere.

Recmo: I think I prefer the synchronous approach, has the least opportunity to screw up.

krzkaczor: Yeah, let's agree on adding async keyword to `increaseTime`. 

Redoing it for sync calls is also a great idea but i don't want to do any more refactoring on this branch. Let's discuss improving things like this tmrw.

test/helpers/chaiWeb3.js
krzkaczor: 👍 i hope one day i will find courage to implement it xD

PR #32, opened 2017-08-22
Move from MiniMeToken to SnapshotToken

Recmo:


test/SnapshotTest.js
```diff
@@ -0,0 +1,93 @@
...
+
+    assert.isFalse(await value.hasValueAt.call(before));
+    assert.isTrue(await value.hasValueAt.call(middle));
+    assert.isTrue(await value.hasValueAt.call(after));
+    assert.equal(41, await value.getValueAt.call(before, 41));
+    assert.equal(100, await value.getValueAt.call(middle, 41));
+    assert.equal(200, await value.getValueAt.call(after, 41));
+  });
+  it("should create daily snapshots", async () => {
+    const day = 24 * 3600;
```
krzkaczor: i would move it outside to the outer scope and declare it only once. Also i refactored it on my branch :D

PR #33, opened 2017-08-23
Use Nefund issued Truffle [Postponed]

Recmo:


mocha.js
```diff
@@ -2,4 +2,12 @@ import chai from "chai";
 import chaiAsPromised from "chai-as-promised";
 import chaiBignumber from "chai-bignumber";
 
-chai.use(chaiAsPromised).use(chaiBignumber(web3.BigNumber));
+const GetNumber = artifacts.require("./test/GetNumber.sol");
+
+before(async () => {
```
krzkaczor: this makes me cry a little bit xD can you left a comment why we invented that thing? for future generations

PR #40, opened 2017-08-25
Reimplemented commitment contracts

krzkaczor:


test/WhitelistedCommitment.js
```diff
@@ -368,19 +399,81 @@ contract(
...
+            fixedTickets: fixedDeclaredTickets,
+            startTimestamp: startingDate,
+            minTicket: etherToWei(0),
+            eurEthRate: etherToWei(0.1) // sets up a rate allowing for rounding errors
+          }
+        });
+
+        const expectedNeumarkAmmount = await curveInEther(
+          fixedDeclaredTickets[0],
+          await commitment.ethEURFraction()
```
krzkaczor: it's the same as `eurEthRate: etherToWei(0.1)` above. Could you extract and reuse that value? I want to avoid getting data from contracts that later we would assert on to minimize the possibility of hitting a bug from contract in test.

rudolfix: ok done

contracts/WhitelistedCommitment.sol
```diff
@@ -89,35 +87,42 @@ contract WhitelistedCommitment is AccessControlled, AccessRoles, PublicCommitmen
...
         // what is above limit for fixed price should be rewarded from curve
         uint256 reward = 0;
-        if ( eth > fixedTicket ) {
-            if (fixedTicket > 0) // recompute euro if part of msg.value goes thru whitelist
-                euros = convertToEUR(eth - fixedTicket);
-            reward = curve.issue(euros); // PublicCommitment.giveNeumarks(investor, eth - fixedTicket, euros);
-            eth = fixedTicket;
+        if ( eth > fixedRemainingTicket ) {
+            if (fixedRemainingTicket > 0) // recompute euro if part of msg.value goes thru whitelist
+                euros = convertToEUR(eth - fixedRemainingTicket);
```
Recmo: Conditionally overwriting a function parameter is very confusing. It's good practice to treat function parameters as constant.

contracts/WhitelistedCommitment.sol
```diff
@@ -89,35 +87,42 @@ contract WhitelistedCommitment is AccessControlled, AccessRoles, PublicCommitmen
...
         }
     }
-    event Debug(uint256 value);
-    /// overrides base class to compute neumark reward for fixed price investors
+
+    /// overrides base class to compute neumark reward for ordered whitelistfix investors
     function giveNeumarks(address investor, uint256 eth, uint256 euros)
         internal
         returns (uint256)
     {
```
Recmo: There should be preconditions like `require(euros == convertToEur(eth))` on start. But really, we should drop the `euros` argument to prevent this class of errors.

Recmo: It looks like the function arguments are redundant. We now have an extra requirement at the start of the function that the eth and euro parameters are correct with each other.

contracts/WhitelistedCommitment.sol
Recmo: Please give all significant intermediate values a descriptive name and try to have one operation per line. In this case,
```
uint256 nonWhitelistEth = eth - fixedRemainingTicket;
euros = convertToEUR(nonWhitelistEth);
```
is way more descriptive.

contracts/WhitelistedCommitment.sol
```diff
@@ -89,35 +87,42 @@ contract WhitelistedCommitment is AccessControlled, AccessRoles, PublicCommitmen
...
-            // rounding errors, send out remainders
-            // @remco review
-            uint256 remainingBalance = neumarkToken.balanceOf(address(this));
-            if (absDiff(fixedreward, remainingBalance) < 1000)
-                fixedreward = remainingBalance; // send all
-            // decrease ticket size
-            fixedCost[investor] -= eth;
+            fixedreward = proportion(fixedRemainingNeumarks, eth, fixedRemainingTicket);
+            // decrease ticket size and neumarks left
+            if ((fixedreward > fixedRemainingNeumarks)
```
Recmo: Please simplify the conditionals, it's hard to tell when something is executed and why. It should be super clear for critical code like this. Feel free to introduce boolean variables to explain things:

```
bool rewardOverflow = fixedreward > fixedRemainingNeumarks;
bool remainderTiny = fixedRemainingNeumarks - fixedreward < 10;
if (rewardOverflow || remainderTiny) {
    …
```

contracts/WhitelistedCommitment.sol
```diff
@@ -89,35 +87,42 @@ contract WhitelistedCommitment is AccessControlled, AccessRoles, PublicCommitmen
...
         if (eth > 0) {
-            fixedreward = proportion(totalFixedCostNeumarks, eth, totalFixedCostAmount);
-            // rounding errors, send out remainders
-            // @remco review
-            uint256 remainingBalance = neumarkToken.balanceOf(address(this));
-            if (absDiff(fixedreward, remainingBalance) < 1000)
-                fixedreward = remainingBalance; // send all
-            // decrease ticket size
-            fixedCost[investor] -= eth;
+            fixedreward = proportion(fixedRemainingNeumarks, eth, fixedRemainingTicket);
```
Recmo: `proportion` should not be called with the second argument larger than the third (we should not have fractions >1 there) because of overflows and stuff.

It's much more elegant here to take
```
uint256 ticketEth = min(eth, fixedRemainingTicket);
```
This prevents overflows and we can elliminate the complex conditional that follows.

contracts/WhitelistedCommitment.sol
```diff
@@ -89,35 +87,42 @@ contract WhitelistedCommitment is AccessControlled, AccessRoles, PublicCommitmen
...
+            // decrease ticket size and neumarks left
+            if ((fixedreward > fixedRemainingNeumarks)
+                || (fixedRemainingNeumarks - fixedreward < 10))
+            {
+                // give rest of the neumarks
+                fixedreward = fixedRemainingNeumarks;
+                // zero whole ticket
+                fixedCostNeumarks[investor] = 0;
+                fixedCostTickets[investor] = 0;
```
Recmo: This case is not necessary.

Recmo: This case can be removed if we use `min`.

contracts/WhitelistedCommitment.sol
rudolfix: this is a good catch! I simply forgot to remove euros parameter

contracts/WhitelistedCommitment.sol
rudolfix: this is not necessary as first we spend the surplus eth with whitelist mechanism and when the proportion is computed eth <= fixedRemainingTicket. Surely some intermediate variable should be added that makes code more readable and this will be done

contracts/WhitelistedCommitment.sol
rudolfix: This case compares Neumarks not eth (I assume you refer to `min` usage below)
```
uint256 ticketEth = min(eth, fixedRemainingTicket);
```
We can remove this case if what you say of your `Disbursal` contract applies here: last proportion is surely 1/1 as all the rounding errors are annihilated before. Let me write some tests for it

contracts/WhitelistedCommitment.sol
```diff
@@ -89,35 +87,42 @@ contract WhitelistedCommitment is AccessControlled, AccessRoles, PublicCommitmen
...
+            } else {
+                fixedCostNeumarks[investor] -= fixedreward;
+                // this will not overflow, we check fixedCostTickets[investor] > eth earlier
+                fixedCostTickets[investor] -= eth;
+            }
```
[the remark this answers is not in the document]
rudolfix: Totally agree that this needs overhaul. Too many changes accumulated. I'll decide if to go with mixin or not, surely base class that will enforce functions to be overriden is a good idea. Thx for the review

contracts/WhitelistedCommitment.sol
rudolfix: Ehhh ok it is trivial to prove that last proportion is 1:1 :) it does not matter at all how you round. you can even randomly round up or down

test/helpers/promiseUtils.js
```diff
@@ -0,0 +1,8 @@
+export async function sequence(array, promiseGenerator) {
+  const results = [];
+  for (const item of array) {
+    results.push(await promiseGenerator(item));
+  }
+
+  return results;
+}
```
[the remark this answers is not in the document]
krzkaczor: Yeah i used to write like this :D but c'mon solution with loop and await is 10x more readable

test/helpers/verification.js
```diff
@@ -52,3 +53,7 @@ export async function curveInEther(money, eurEtherRatio) {
 export function ethToEur(ether, eurEtherRatio = etherToWei(218.1192809)) {
```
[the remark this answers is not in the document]
rudolfix: All test helpers, including `verification.js` need overhaul. it is pointless to just fix this issue. it would require to move `eurToEth` to `deploy.js` where deployment config values are available so it is substantial amount of work and I doubt it will be final structure when tests are done correctly. In short: can we merge solidity and do testing improvements in separate PR?

contracts/AccessRoles.sol
```diff
@@ -4,6 +4,12 @@ contract AccessRoles {
...
     // may setup LockedAccount, change disbursal mechanism and set migration
     bytes32 public constant ROLE_LOCKED_ACCOUNT_ADMIN = keccak256("LockedAccountAdmin");
+
     // may setup whitelists and abort whitelisting contract with curve rollback
     bytes32 public constant ROLE_WHITELIST_ADMIN = keccak256("WhitelistAdmin");
+
+    bytes32 public constant ROLE_NEUMARK_ISSUER = keccak256("NeumarkIssuer");
+    bytes32 public constant ROLE_NEUMARK_BURNER = keccak256("NeumarkBurner");
+    bytes32 public constant ROLE_TRANSFERS_ADMIN = keccak256("TransferAdmin");
+    bytes32 public constant ROLE_SNAPSHOT_CREATOR = keccak256("SnapshotCreator");
```
Recmo: These are changes already incorporated in master. I also see commits from master in the PR. Did something go wrong during a rebase?

Recmo: Older version of `rb/curve` branch got merged in. It seems okay.

PR #42, opened 2017-08-25
replace ownable with AccessControl in LockedAccount and WhitelistedCommitment

rudolfix:


contracts/AccessControl/RoleBasedAccessControl.sol
```diff
@@ -97,7 +97,7 @@ contract RoleBasedAccessControl is IAccessPolicy, AccessControlled {
             bool grantedLocal = localAccess == TriState.True;
 
             // Log and return
-            Access(subject, role, object, verb, grantedLocal);
+            Access(subject, role, object, verb, granted);
```
Recmo: Where does `granted` come from?

contracts/WhitelistedCommitment.sol
```diff
@@ -73,11 +73,11 @@ contract WhitelistedCommitment is PublicCommitment {
     // @remco this is a small breach of trust as we can invalidate terms any moment
     function abortCommitment()
         public
-        onlyOwner
+        only(ROLE_WHITELIST_ADMIN)
     {
         require(currentTime()<startDate);
         rollbackCurve();
-        selfdestruct(owner);
+        selfdestruct(address(0));
```
Recmo: You can also use `msg.sender`.

contracts/WhitelistedCommitment.sol
```diff
@@ -133,9 +133,11 @@ contract WhitelistedCommitment is PublicCommitment {
     }
 
     function WhitelistedCommitment(uint256 _startDate, uint256 _endDate, uint256 _minCommitment, uint256 _maxCommitment,
-        uint256 _minTicket, uint256 _ethEurFraction, ITokenWithDeposit _ethToken, LockedAccount _lockedAccount, Curve _curve)
+        uint256 _minTicket, uint256 _ethEurFraction, ITokenWithDeposit _ethToken, LockedAccount _lockedAccount,
+        Curve _curve, IAccessPolicy _policy)
```
Recmo: Convention elsewhere is to put forwarded constructor arguments in the order of the base classes, in  this case `_policy` would come first.

contracts/test/TestLockedAccountMigrationTarget.sol
```diff
@@ -52,9 +52,9 @@ contract TestLockedAccountMigrationTarget is LockedAccount, LockedAccountMigrati
         return true;
     }
 
-    function TestLockedAccountMigrationTarget(IERC667Token _assetToken, Curve _neumarkCurve,
+    function TestLockedAccountMigrationTarget(IERC667Token _assetToken, Curve _neumarkCurve, IAccessPolicy _policy,
```
Recmo: Convention elsewhere is to put forwarded arguments first and in the order of the base classes. `_policy` would come first.

migrations/2_deploy_contracts.js
```diff
@@ -51,10 +57,26 @@ module.exports = deployer =>
...
-    await lock.setController(publicCommitment.address);
+    console.log("Seting permissions");
+    await deployer.deploy(AccessRoles);
+    const accessRoles = await AccessRoles.deployed();
+    await accessControl.setUserRole(
+      accounts[1],
+      await accessRoles.ROLE_LOCKED_ACCOUNT_ADMIN(),
+      lock.address,
+      1
+    ); // 1 is True
```
Recmo: I would put a `const TriState = { Unset:0, Allowed: 1, Denied: 2};` at the top of this file and use that. Than we don't need the comment, don't have magic numbers and can change it in one location.

Even nicer would be to get the enum values directly out of the contract, but that may not be worth the effort.

test/AccessControl.js
```diff
@@ -0,0 +1,109 @@
...
+  beforeEach(async () => {
+    accessControl = await RoleBasedAccessControl.new();
+    accessControlled = await TestAccessControlTruffleMixin.new(
+      accessControl.address
+    );
+
+    exampleRole = await accessControlled.ROLE_EXAMPLE();
+  });
+
+  function expectAccessChangedEvent(
```
Recmo: This can be generalized to any event and put in helpers. But let's do that in a separate PR.

test/AccessControl.js
```diff
@@ -0,0 +1,109 @@
...
+      accessControlled.address,
+      1
+    );
+    expectAccessChangedEvent(
+      tx,
+      owner1,
+      exampleRole,
+      accessControlled.address,
+      0,
+      1
```
Recmo: `TriState.*`.

contracts/AccessControl/RoleBasedAccessControl.sol
Recmo: In PR #39 this is refactored, this will result in a merge conflict. When this PR is merged, I can rebase #39 on top of it.

rudolfix: my fault when putting parameters back into order! there is no error for unknown variable or no warning for using uninitialized one, so it is a little weird it compiles.

migrations/2_deploy_contracts.js
krzkaczor: 👍 for extracting magic numbers to top of the file

test/helpers/deploy.js
```diff
@@ -9,9 +9,12 @@ const Neumark = artifacts.require("Neumark");
 const Curve = artifacts.require("Curve");
 const TestCommitment = artifacts.require("TestCommitment");
 const WhitelistedCommitment = artifacts.require("WhitelistedCommitment");
-const FeeDistributionPool = artifacts.require("FeeDistributionPool");
+const RoleBasedAccessControl = artifacts.require("RoleBasedAccessControl");
+const AccessRoles = artifacts.require("AccessRoles");
 
 export async function deployAllContracts(
+  lockAdminAccount,
```
krzkaczor: make it part of the config object (last argument)

test/helpers/deploy.js
```diff
@@ -9,9 +9,12 @@ const Neumark = artifacts.require("Neumark");
...
+  whitelistAdminAccount,
```
krzkaczor: same

contracts/AccessControl/RoleBasedAccessControl.sol
Recmo: Solidity allocates all local variables of a scope at the start of the scope, `bool granted` is declared later in this scope. That's why it compiles. I think it shouldn't though. The value will be undefined.

test/helpers/deploy.js
rudolfix: why? I need to send those with each call, they are not part of configuration but runtime

test/WhitelistedCommitment.js
```diff
@@ -15,598 +15,699 @@ import {
...
-      await expect(commitment.setFixed(investors, tickets)).to.be.rejectedWith(
-        EvmError
-      );
-    });
-
-    it("should not be possible to set it after commitment is started", async () => {
-      const startingDate = closeFutureDate();
-      const { commitment, curve } = await deployAllContracts({
-        commitmentCfg: { startTimestamp: startingDate }
+contract(
```
krzkaczor: @Recmo that's why i wanted to have prettify working with 100 chars in a line :D

test/helpers/deploy.js
krzkaczor: Ohh shit. Github hid some files from me and i thought that this is somehow optional. Can you leave then 'invariant' call at the beginning of the function to throw if required args are not provided? example: https://github.com/Neufund/ico-contracts/blob/master/test/helpers/deploy.js#L61

rudolfix: sure

PR #43, opened 2017-08-27
Eslint fixes

krzkaczor:
This brings error no from ~150 to ~50. I turned off some rules that i found not really useful.


test/Curve.js
```diff
@@ -147,26 +143,27 @@ contract("Curve", accounts => {
...
   });
 
   it("should issue Neumarks", async () => {
     assert.equal((await curve.totalEuroUlps.call()).valueOf(), 0);
     assert.equal((await neumark.totalSupply.call()).valueOf(), 0);
 
-    const r1 = await curve.issue(EUR_DECIMALS.mul(100), { from: accounts[1] }); // TODO check result
-    console.log(`\tIssue took ${gasCost(r1)}.`);
+    const r1 = await curve.issue(EUR_DECIMALS.mul(100), { from: accounts[1] });
+    expect(txGasCost(r1)).to.be.eq(191726);
```
Recmo: I like to have gas cost of tests as part of the test output, it is an import resource to be aware off. Compare it with the running time that mocha already prints.

test/Neumark.js
```diff
@@ -13,10 +11,6 @@ contract("Neumark", accounts => {
     await neumark.changeController(controller.address);
   });
 
-  it("should deploy", async () => {
-    console.log(`\tNeumark took ${gasCost(neumark)}.`);
-    console.log(`\tController took ${gasCost(controller)}.`);
-  });
```
Recmo: Gas cost of contract deployment is also an important resource. It is possible to write contracts that are to big to deploy.

test/SnapshotTest.js
```diff
@@ -22,10 +23,6 @@ contract("Snapshot", accounts => {
     value = await SnapshotTest.new();
   });
 
-  it("should deploy", async () => {
-    console.log(`\tSnapshot took ${gasCost(value)}.`);
-  });
-
```
Recmo: Same here.

test/helpers/chaiHelpers.js
```diff
@@ -1,11 +1,14 @@
...
+export default function(chai) {
+  chai.Assertion.addMethod("balanceWith", function balanceWith({
+    ether,
+    neumarks
+  }) {
     invariant(ether, "missing ether parameter");
     invariant(neumarks, "missing neumarks parameter");
 
-    const balance = this._obj;
+    const balance = this._obj; // eslint-disable-line no-underscore-dangle
```
Recmo: Use `// eslint-disable-next-line` Looks much better

test/helpers/chaiWeb3.js
```diff
@@ -1,26 +1,29 @@
...
-  chai.Assertion.addMethod("blockchainArrayOfSize", async function(size) {
-    // I would love to hear ideas for better implementation
+export default function(chai) {
+  chai.Assertion.addMethod(
+    "blockchainArrayOfSize",
+    async function blockchainArrayOfSize(size) {
+      // I would love to hear ideas for better implementation
 
-    const web3ArrayAccessor = this._obj;
+      const web3ArrayAccessor = this._obj; // eslint-disable-line no-underscore-dangle
```
Recmo: same here.

test/helpers/gasCost.js
```diff
@@ -7,12 +9,13 @@ export const gasCost = gas =>
     1000 * gas / gasLimit
   ) / 10}% of limit)`;
 
-export const txGasCost = tx => gasCost(tx.receipt.gasUsed);
+export const txGasCost = tx => {
+  invariant(
```
Recmo: Why not `expect(tx).to.have.property('receipt')` ?

krzkaczor: I use `invariant` as a way for checking for developer errors and expect is for test expections. I can unify it, I don't really care :D

Recmo: Me neither. Just curious about it.

PR #44, opened 2017-08-28
Add Reclaim contract

Recmo:


contracts/Commitment/WhitelistedCommitment.sol
```diff
@@ -1,8 +1,10 @@
 pragma solidity 0.4.15;
 
 import './PublicCommitment.sol';
+import '../Reclaimable.sol';
 
-contract WhitelistedCommitment is AccessControlled, AccessRoles, CommitmentBase {
+
+contract WhitelistedCommitment is AccessControlled, AccessRoles, CommitmentBase, Reclaimable {
```
rudolfix: It should not allow to reclaim `Neumark` token before WhitelistedCommitment is finalized,  Neumarks are already handled differently
* `abortCommitment()` will burn all neumarks in possession of the contract if called before `startDate`
* no one should be able to reclaim or burn Neumarks when commitment is in progress

use `isFinalized()` function to get finalization status
All neumarks in contract possession after finalization can be reclaimed

contracts/Commitment/PublicCommitment.sol
```diff
@@ -43,11 +44,14 @@ contract PublicCommitment is CommitmentBase {
     }
 
     function PublicCommitment(
+        IAccessPolicy _policy,
```
rudolfix: You may move it to `BaseCommitment` class in that case, it seems that `AccessControlled` and `Reclaimable` are obligatory for any commitment contract.

test/Reclaimable.js
```diff
@@ -0,0 +1,58 @@
...
+    expect(reclaimerAfter.sub(reclaimerBefore)).to.be.bignumber.equal(amount);
+  });
+  it("should only allow ROLE_RECLAIMER to reclaim", async () => {
+    const amount = web3.toWei(1, "ether");
+    await forceEther(reclaimable.address, amount);
+    await expectThrow(reclaimable.reclaim(RECLAIM_ETHER, { from: other }));
+    await expectThrow(reclaimable.reclaim(RECLAIM_ETHER, { from: deployer }));
+    await reclaimable.reclaim(RECLAIM_ETHER, { from: reclaimer });
+  });
+});
```
rudolfix: I would add following pending tests, we should implement them later
```
it("should not reclaim Neumark token before WhitelistedCommitment is finalized");
it("should not be able to reclaim ether from EtherToken");
it("should not be able to reclaim assetToken from LockedAccount");
```

Recmo: Added to the respective tests

contracts/Commitment/PublicCommitment.sol
Recmo: Done.

PR #46, opened 2017-08-28
Fix reentrancy issue (critical)

Recmo:


contracts/EtherToken.sol
```diff
@@ -49,9 +49,9 @@ contract EtherToken is StandardToken, ITokenWithDeposit {
         public
     {
         require(balances[msg.sender] >= amount);
-        assert(msg.sender.send(amount));
         balances[msg.sender] -= amount;
         totalSupply -= amount;
+        assert(msg.sender.send(amount));
```
rudolfix: true but purely theoretical. the gas limit is set to minimum for send so any inter-contract call is not possible anymore

rudolfix: to test it you need to change `send` intro `call` like
```
msg.sender.call.value(amount)()
```

PR #49, opened 2017-08-29
Add EthereumForkArbiter

Recmo:


test/EthereumForkArbiter.js
```diff
@@ -0,0 +1,119 @@
...
+    await ethereumForkArbiter.signFork(block.number, block.hash, {
+      from: arbiter
+    });
+  });
+  it("should remember last signed fork", async () => {
+    const block = await web3.eth.getBlock("latest");
+    const tx = await ethereumForkArbiter.signFork(block.number, block.hash, {
+      from: arbiter
+    });
+    const rNumber = await ethereumForkArbiter.lastSignedBlockNumber.call();
```
krzkaczor: is `r` coming from Remco? xD I think that standard testing vocabulary is to use "expected" for expected values and actual for ones coming from tested code.

test/EthereumForkArbiter.js
```diff
@@ -0,0 +1,119 @@
...
+    const rHash = await ethereumForkArbiter.lastSignedBlockHash.call();
+    const rTime = await ethereumForkArbiter.lastSignedTimestamp.call();
+    expect(rNumber).to.be.bignumber.equal(block.number);
+    expect(rHash).to.equal(block.hash);
+    const rBlock = await web3.eth.getBlock(tx.receipt.blockNumber);
```
krzkaczor: how about following 3A principle? So do not interleave asserts with anything else?

test/EthereumForkArbiter.js
```diff
@@ -0,0 +1,119 @@
...
+        from: other
+      })
+    );
+    await ethereumForkArbiter.announceFork(name, url, {
+      from: arbiter
+    });
+  });
+  it("should only allow ROLE_FORK_ARBITER to sign", async () => {
+    const block = await web3.eth.getBlock("latest");
+    await expectThrow(
```
krzkaczor: honestly, i would prefer to stick to expect + promised chai so something like:
`await expect(something).to.be.rejectedWith(EvmError); `

Recmo: I like that too, but I want some abstraction over the specifics of the EVM error. It tends to change with clients and versions, check the `expectThrow` implementation to see what I mean.

test/EthereumForkArbiter.js
Recmo: Co to jest 3A principle?

test/EthereumForkArbiter.js
Recmo: “Received”. Will change to expected/actual

test/EthereumForkArbiter.js
```diff
@@ -0,0 +1,119 @@
...
+contract("EthereumForkArbiter", ([deployer, arbiter, other]) => {
+  let ethereumForkArbiter;
+
+  beforeEach(async () => {
+    const accessPolicy = await createAccessPolicy([
+      { subject: arbiter, role: "ROLE_FORK_ARBITER" }
+    ]);
+    ethereumForkArbiter = await EthereumForkArbiter.new(accessPolicy);
+  });
+  it("should deploy", async () => {
```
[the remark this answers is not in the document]
Recmo: Deblob the code. Roger.

contracts/EthereumForkArbiter.sol
```diff
@@ -0,0 +1,53 @@
...
+        public
+        only(ROLE_FORK_ARBITER)
+    {
+        require(block.blockhash(number) == hash);
+        lastSignedBlockNumber = number;
+        lastSignedBlockHash = hash;
+        lastSignedTimestamp = block.timestamp;
+        ForkSigned(lastSignedBlockNumber, lastSignedBlockHash);
+    }
+}
```
rudolfix: Conceptual problem: how this is intended to be used? Should we aggregate arbiter in each contract we care about forks? Like Neumark and LockedAccount?

Recmo: Just one globally should be enough, that's how I deploy it.

Recmo: Multiple would be a bad idea, they could only conflict.

Recmo: They don't need to be linked to the contracts. I did not want to put in a `kill all bad forks` mechanism for security reasons.

Only thing we need to do is put the contract address in the agreements in the paragraph that says how to identify the one-true-fork.

rudolfix: Yes, there should be just one, what I meant is that you keep its address in any contract that has agreement attached. In that case you have the same mechanism for legal and smart contract - a link to fork arbiter... We may think on adding arbiter with mixin that adds url to legal agreement. Let's discuss tomorrow

Recmo: I like the idea of connecting contracts back to the ForkArbiter. It also seems to overlap with the Agreement contract. I want to combine the two, so let's get this PR in and finish it in Agreement one, ok?

PR #51, opened 2017-08-29
Add Agreement contract

Recmo:


contracts/LockedAccount.sol
```diff
@@ -144,6 +156,7 @@ contract LockedAccount is AccessControlled, AccessRoles, TimeSource, ReturnsErro
     // if used before longstop date, calculates penalty and distributes it as revenue
     function unlock()
         onlyStates(LockState.AcceptingUnlocks, LockState.ReleaseAll)
+        acceptAgreement(msg.sender)
```
rudolfix: Each investor accepts agreement by calling lock(). We can safely remove it here

contracts/LockedAccount.sol
```diff
@@ -153,6 +166,7 @@ contract LockedAccount is AccessControlled, AccessRoles, TimeSource, ReturnsErro
     // this allows to unlock and allow neumarks to be burned in one transaction
     function receiveApproval(address from, uint256 _amount, address _token, bytes _data)
         onlyStates(LockState.AcceptingUnlocks, LockState.ReleaseAll)
+        acceptAgreement(from)
```
rudolfix: Each investor accepts agreement by calling lock(). We can safely remove it here

contracts/LockedAccount.sol
```diff
@@ -210,6 +224,7 @@ contract LockedAccount is AccessControlled, AccessRoles, TimeSource, ReturnsErro
 
     /// migrate single investor
     function migrate()
+        acceptAgreement(msg.sender)
```
rudolfix: `migrate()` results in signing agreement in `migration` contract so we should remove it here. migration event is generated anyway

contracts/Neumark.sol
```diff
@@ -51,6 +56,7 @@ contract Neumark is
     function issueForEuro(uint256 euroUlps)
         public
         only(ROLE_NEUMARK_ISSUER)
+        acceptAgreement(msg.sender)
```
rudolfix: please keep in mind that currently it is called by Commitment smart contract only

contracts/Neumark.sol
```diff
@@ -68,6 +74,7 @@ contract Neumark is
     function burnNeumark(uint256 neumarkUlps)
         public
         only(ROLE_NEUMARK_BURNER)
+        acceptAgreement(msg.sender)
```
rudolfix: please keep in mind that currently it is called by LockedAccount contract only

contracts/Neumark.sol
```diff
@@ -125,6 +134,8 @@ contract Neumark is
         uint amount
     )
         internal
+        acceptAgreement(owner)
+        acceptAgreement(spender)
```
rudolfix: we for sure should accept for `owner`. I would refrain from accepting in the name of `spender` as spender may not have a balance >0  as a result. he may be a broker.
I think we should ask Jacek. it may be we want also to include brokers in here.

contracts/Neumark.sol
```diff
@@ -108,6 +115,8 @@ contract Neumark is
         uint amount
     )
         internal
+        acceptAgreement(from)
+        acceptAgreement(to)
```
rudolfix: We'll be generating quite a lot of events here. What is our gas cost for simple transfer of Neumarks?

contracts/LockedAccount.sol
Recmo: I was errering on the side of safety because i wasn't sure. Removed it now.

contracts/LockedAccount.sol
Recmo: ok.

contracts/LockedAccount.sol
Recmo: ok/

contracts/Neumark.sol
Recmo: Yes, but this might change in the future. Jacek recommended to issue the event whenever someone holds NMK, which this function does.

contracts/Neumark.sol
Recmo: I'm aware. This might change in the future, so we need to have it in. It might be redundant with the `mOnTransfer` one though. I'll check that.

contracts/Neumark.sol
Recmo: I'll benchmark this. LogEvents are pretty cheap, about 500 gas or so (+ 400 for each index).

contracts/Neumark.sol
Recmo: True. This may be redundant. The allowance case has not been discussed with Jacek.

contracts/Neumark.sol
Recmo: With both log events:

```
        Transfer took 81834 gas (€0.52, 1.2% of limit)
```

without log events:

```
        Transfer took 79368 gas (€0.5, 1.2% of limit)
```

So with overhead, the two events cost us `2466` gas.

rudolfix: thanks!

contracts/Agreement.sol
```diff
@@ -0,0 +1,23 @@
...
+
+
+contract Agreement {
+
+    IEthereumForkArbiter public ethereumForkArbiter;
+    string public agreementUri;
+
+    event AgreementAccepted(address indexed accepter);
+
+    modifier acceptAgreement(address accepter) {
```
rudolfix: One conceptual issue: should we have a generic `acceptAgreement()` method that allows to explicitly sign this agreement? 
OK we should not - if it is necessary it should be implemented in derived contract. Ideally acceptance should be a part of a transaction that has some substance to it like `lock` or `transfer`
Just leaving this rant here so I do not forget it and not raise it back :>

Recmo: I was thinking the same thing, have a modifier `implicitAccept` and a function `explicitAccept`. Note that current `acceptAgreement` does not even require the accepter to sign or even be involved.
We could also maintain state of who accepted and who didn't. At the moment all of this doesn't seem worth the complexity.

PR #57, opened 2017-08-30
more LockedAccount tests

rudolfix:
1. LockedAccount tests
2. Reclaimable allows to reclaim part of the available balance (allows to reclaim surplus ether from etherToken and surplus etherToken from LockedAccount)
3. Tests for above


rudolfix:
@Recmo do you think those changes to reclaimable do not make sense? there's some merit in extracting ether above totalSupply from ether token.
PR is different thing it can be done

PR #59, opened 2017-08-31
Add ERC223 to SnapshotToken, fix ERC677 naming  and clean up.

Recmo:


contracts/SnapshotToken/Helpers/TokenMetadata.sol
```diff
@@ -0,0 +1,55 @@
...
+        string _tokenSymbol,
+        string _version
+    ) {
+        tokenName = _tokenName;                                 // Set the name
+        tokenDecimals = _decimalUnits;                          // Set the decimals
+        tokenSymbol = _tokenSymbol;                             // Set the symbol
+        tokenVersion = _version;
+    }
+
+    function symbol()
```
rudolfix: This contract has two sets of functions returning same value. There's `symbol()` and there's `tokenSymbol()` which compiler generates. Why so?

contracts/SnapshotToken/SnapshotToken.sol
```diff
@@ -1,15 +1,18 @@
 pragma solidity 0.4.15;
 
 import '../Snapshot/DailyAndSnapshotable.sol';
-import '../Standards/IERC667Token.sol';
-import '../Standards/IERC667Callback.sol';
+import '../Standards/IERC223Token.sol';
+import '../Standards/IERC223Callback.sol';
+import '../Standards/IERC677Token.sol';
+import '../Standards/IERC677Callback.sol';
```
rudolfix: Do we still plan to support both? Didn't we decide to keep ERC223?

contracts/SnapshotToken/SnapshotToken.sol
```diff
@@ -90,9 +95,30 @@ contract SnapshotToken is
...
+        if (!success) {
+            return success;
+        }
+
+        // Notify the receiving contract.
+        if (isContract(to)) {
+            IERC223Callback(to).tokenFallback(msg.sender, amount, data);
+        }
+
+        Transfer(msg.sender, to, amount, data);
```
rudolfix: I do not like the idea to have two Transfer events. `data` is not affecting token contract in any way so why log it? If it has any impact in fallback, the receiving contract can log accordingly. Standard has just one event but it is a breaking change and I do not like it even more.

contracts/SnapshotToken/Helpers/TokenMetadata.sol
rudolfix: Ok got it, you want interface. If you use `string public token;` it will not auto-implement by compiler ;>?
Should at least `string public tokenName; ` be internal to not expose same data twice?

Recmo: > If you use string public token; it will not auto-implement by compiler ;>?

It will, but Solidity does not count it as implementing the interface, so it will think the contract is abstract and not produce code, but also not fail ( == stupid).

You are right, I should make it  private.

contracts/SnapshotToken/SnapshotToken.sol
Recmo: With this PR it will support both. We don't have to use both though. The overhead for supporting ER677 is low. (Unless you want to drop the entire Allowance mechanis and go pure ERC223, but than we loose ERC20 support :) )

contracts/SnapshotToken/SnapshotToken.sol
Recmo: I also don't like it for exactly the reasons you mention, but ERC223 requires it to have data in there. I don't think it will be a huge issue if we also break away from ERC223 on this one. Shall I remove the extra event?

contracts/SnapshotToken/SnapshotToken.sol
rudolfix: OK agree

PR #60, opened 2017-08-31
Cleanup roles

Recmo:


test/helpers/verification.js
```diff
@@ -20,19 +18,19 @@ async function deployNeumark() {
   // TODO: more specific rights
   await rbac.setUserRole(
     EVERYONE,
-    await roles.ROLE_NEUMARK_ISSUER(),
+    web3.sha3("NeumarkIssuer"),
```
krzkaczor: what about creating roles module with all roles exported as hashes already? (like a object NAME->HASH)

Recmo: Good idea. I already had random breakage because typos :D

migrations/2_deploy_contracts.js
```diff
@@ -75,35 +74,33 @@ module.exports = function deployContracts(deployer, network, accounts) {
     );
     console.log("Commitment terms set");
     console.log("Seting permissions");
-    await deployer.deploy(AccessRoles);
-    const accessRoles = await AccessRoles.deployed();
     await accessControl.setUserRole(
       publicCommitment.address,
-      await accessRoles.ROLE_NEUMARK_ISSUER(),
+      web3.sha3("NeumarkIssuer"),
```
krzkaczor: You don't want to reuse "roles" module here?

Recmo: No, don't want to depend on a folder called `test` in a file called `deploy`. Moving it somewhere else also doesn't seem elegant. I think it's good if the deploy script is very concrete.

Recmo: It would make more sense to have a folder of deployment scripts, and re-use those in test.

krzkaczor: Yeah, I think that best solution would be to have a separate place for these "common" constants, utils whatever. I am worried that we will make a typo or something like this in deployment scripts. I think we should address that before prod deploy.  For now i can accept that.

PR #61, opened 2017-08-31
Successful whitelisting tests

krzkaczor:


contracts/Commitment/WhitelistedCommitment.sol
```diff
@@ -139,6 +139,7 @@ contract WhitelistedCommitment is AccessRoles, CommitmentBase {
 
     function reclaim(IBasicToken token)
         public
+        only(ROLE_RECLAIMER)
```
Recmo: Isn't this redundant with the check that already happens in `Reclaimable.reclaim` ?

krzkaczor: Yeah, it is. I am not sure what would be our general strategy for such situations? 

I would argue to leave it here because in future someone can modify this method (call another contract or something) assuming that the only reclaimer could do it but in a reality, everyone can.

Recmo: Agreed. I like the added readability. The extra gas cost is neglible since this is an 'admin' call anyway.

For sake of consistency we should then do the same on `LockedAccount` and `EtherToken`.

test/WhitelistedCommitment.js
```diff
@@ -726,15 +726,64 @@ contract(
...
+          }
+        });
+        await setTimeTo(startingDate);
+        await commitment.commit({
+          value: actualInvestor1Commitment,
+          from: investor1
+        });
+
+        await setTimeTo(startingDate + duration + HOUR);
+
```
rudolfix: where do you finalize commitment? the point of this test is to make sure that Neumarks are not transferable after finalization. Which differs for the public commitment which enables Neumarks

PR #63, opened 2017-08-31
[Idea] [WIP] Introduce ROLE_NEUMARK_TRADER

Recmo:
Here's an idea. We introduce a role NEUMARK_TRADER for the transfer function. We can then remove all of the transferEnabled / transferAdmin stuff and instead control it by setting (EVERYONE, ROLE_NEUMARK_TRADER, NEUMARK, Allow/Deny). We can give Commitment contracts special permission, so we can also remove the enableTransfer stuff there.

It both increases functionality and reduces code complexity. Only downside is that there won't be a TransferAdmin anymore, but intead this needs to be the AccessController.

(which can always be changed later by upgrading the accesspolicy)

### DIFF
--- a/contracts/Commitment/CommitmentBase.sol
+++ b/contracts/Commitment/CommitmentBase.sol
@@ -162,13 +162,8 @@ contract CommitmentBase is AccessControlled, TimeSource, Math, ITokenOffering, R
     {
         // distribute half half
         uint256 investorNeumarks = divRound(neumarks, neumarkRewardPlatformOperatorDivisor);
-        // @ remco is there a better way to distribute?
-        bool isEnabled = neumark.transferEnabled();
-        if (!isEnabled)
-            neumark.enableTransfer(true);
         require(neumark.transfer(investor, investorNeumarks));
         require(neumark.transfer(platformOperatorWallet, neumarks - investorNeumarks));
-        neumark.enableTransfer(isEnabled);
         return investorNeumarks;
     }
 
--- a/contracts/Commitment/PublicCommitment.sol
+++ b/contracts/Commitment/PublicCommitment.sol
@@ -1,6 +1,7 @@
 pragma solidity 0.4.15;
 
 import "./CommitmentBase.sol";
+import "../AccessControl/RoleBasedAccessControl.sol";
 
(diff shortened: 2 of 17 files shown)


contracts/Commitment/PublicCommitment.sol
```diff
@@ -13,6 +14,13 @@ contract PublicCommitment is CommitmentBase {
     {
         // enable Neumark trading in token controller
         neumark.enableTransfer(true);
+        RoleBasedAccessControl(neumark.accessPolicy()).setUserRole(
+            RoleBasedAccessControl.EVERYONE,
+            ROLE_NEUMARK_TRADER,
+            neumark,
+            RoleBasedAccessControl.TriState.Allow
+        );
+
```
Recmo: This would require PublicCommitment to have ROLE_ACCESS_CONTROLLER. I don't like that.

contracts/Commitment/CommitmentBase.sol
```diff
@@ -162,13 +162,8 @@ contract CommitmentBase is AccessControlled, TimeSource, Math, ITokenOffering, R
     {
         // distribute half half
         uint256 investorNeumarks = divRound(neumarks, neumarkRewardPlatformOperatorDivisor);
-        // @ remco is there a better way to distribute?
-        bool isEnabled = neumark.transferEnabled();
-        if (!isEnabled)
-            neumark.enableTransfer(true);
         require(neumark.transfer(investor, investorNeumarks));
         require(neumark.transfer(platformOperatorWallet, neumarks - investorNeumarks));
```
Recmo: CommitBase will always have ROLE_NEUMARK_TRADER, this is the cool thing.

contracts/Commitment/CommitmentBase.sol
```diff
@@ -162,13 +162,8 @@ contract CommitmentBase is AccessControlled, TimeSource, Math, ITokenOffering, R
     {
         // distribute half half
         uint256 investorNeumarks = divRound(neumarks, neumarkRewardPlatformOperatorDivisor);
-        // @ remco is there a better way to distribute?
```
rudolfix: I love this part

PR #65, opened 2017-08-31
Improve performance of CI

Recmo:
## CI

Starting performance: 22 min 20 sec (7 min 27 sec till test starts)

## Local

Non-instrumented: 1:27.20 (was 2:57.08)

Instrumented: (was 11:07.10)

### DIFF
--- a/test/AccessControl.js
+++ b/test/AccessControl.js
@@ -3,6 +3,7 @@ import EvmError from "./helpers/EVMThrow";
 import { eventValue } from "./helpers/events";
 import { TriState } from "./helpers/triState";
 import roles from "./helpers/roles";
+import { saveBlockchain, restoreBlockchain } from "./helpers/evmCommands";
 
 const RoleBasedAccessControl = artifacts.require("RoleBasedAccessControl");
 const TestAccessControlTruffleMixin = artifacts.require(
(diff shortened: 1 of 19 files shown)


truffle.js
```diff
@@ -1,12 +1,12 @@
...
+const TestRPC = require("ethereumjs-testrpc");
 
 module.exports = {
   networks: {
     development: {
-      host: "localhost",
-      port: 8545,
-      network_id: "*" // Match any network id
+      network_id: "*",
+      provider: TestRPC.provider()
```
Recmo: This helps a lot. Eliminates all (slow) http traffic during test. Unfortunately does not get used when runing solcover.

test/AccessControl.js
```diff
@@ -3,24 +3,32 @@ import EvmError from "./helpers/EVMThrow";
...
       accessControl.address
     );
 
     exampleRole = roles.example;
+    snapshot = await saveBlockchain();
+  });
+
+  beforeEach(async () => {
+    await restoreBlockchain(snapshot);
+    snapshot = await saveBlockchain();
```
krzkaczor: Is this because you cannot restore twice the same snapshot? weeeiiirrddd

test/helpers/evmCommands.js
```diff
@@ -0,0 +1,17 @@
+export const promisify = func => async (...args) =>
+  new Promise((accept, reject) =>
+    func(...args, (error, result) => (error ? reject(error) : accept(result)))
+  );
+
+export const rpcCommand = method => async (...params) =>
+  (await promisify(web3.currentProvider.sendAsync)({
```
krzkaczor: why did you decided to use async?

test/AccessControl.js
Recmo: Yeah, seems so. I think it's more of a rollback operation than a restore.

PR #66, opened 2017-09-04
Enforce solidity linting

Recmo:


Recmo:
I'm aware of the following output:

```
[DEPRECATED] You are using a deprecated soliumrc configuration format. Please see the documentation to migrate from Solium v0 to v1.

[DEPRECATED] Rule "double-quotes" is deprecated. Please use "quotes" instead.
```

Unfortunately, there is no easy way to fix it. The documentation is a lie. Re-initializing the project to generate a fresh new config file results in Solium crashing on start. Honestly, the added value of Solium is not enough to spend too much time on it. It mostly checks white space and simple stuff.

PR #67, opened 2017-09-04
Verify callback caller

Recmo:
Currently `LockedAccount` has an issue where in the state `ReleaseAll` it is possible to unlock for anyone (not just `msg.sender` as in `unlock()`) by calling the `receiveApproval` function. The root cause is that it does not verify being called by the token contract (trusting parameters instead). This PR adds explicit checks for callback origins.

It seems to have a minor impact but is certainly unintended behaviour. It effectively makes `unlockFor` a public function.

### DIFF
--- a/contracts/LockedAccount.sol
+++ b/contracts/LockedAccount.sol
@@ -170,6 +170,7 @@ contract LockedAccount is
         public
         returns (bool)
     {
+        require(msg.sender == _token);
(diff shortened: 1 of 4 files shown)


rudolfix:
OK. I had idea for additional test but when we port to ERC223 fallback, there's no token address being send  so we cannot test against token lying about its address.

PR #72, opened 2017-09-07
adds euro token

rudolfix:
This PR proposes:
1. EUR-T implementation
2. Abstraction on migration pattern that if reviewed OK will be also applied on LockedAccount and is intended to be used on other contract in the future

There are no test. Please review if this goes in right direction regarding the coding pattern and style. If so I will write required tests

### DIFF
--- a/contracts/EuroToken.sol
+++ b/contracts/EuroToken.sol
@@ -86,6 +86,7 @@ contract EuroToken is
         AccessControlled(accessPolicy)
         StandardToken()
         TokenMetadata(NAME, DECIMALS, SYMBOL, "")
+        MigrationSource(accessPolicy, ROLE_EURT_DEPOSIT_MANAGER)
         Reclaimable()
     {
     }
@@ -200,14 +201,6 @@ contract EuroToken is
         require(success);
         // set event
         OwnerMigrated(msg.sender, amount);
-    }
-
-    function enableMigration(IMigrationTarget migration)
-        public
-        only(ROLE_EURT_DEPOSIT_MANAGER)
-        onlyMigrationEnabledOnce
-    {
-        return MigrationSource.enableMigration(migration);
     }
 
     //
--- a/contracts/EuroToken.sol
+++ b/contracts/EuroToken.sol
@@ -1,0 +1,242 @@
+pragma solidity 0.4.15;
+
+import './AccessControl/AccessControlled.sol';
+import './Math.sol';
+import './Reclaimable.sol';
+import './Standards/IERC677Token.sol';
+import './Standards/IERC677Callback.sol';
+import './Standards/ITokenWithDeposit.sol';
+import './SnapshotToken/Helpers/TokenMetadata.sol';
+import './Zeppelin/StandardToken.sol';
+import './MigrationSource.sol';
+import './EuroTokenMigrationTarget.sol';
+
+
+/// Simple implementation of EuroToken which is pegged 1:1 to certain off-chain
+/// pool of Euro. Balances of this token are intended to be migrated to final
+/// implementation that will be available later
+contract EuroToken is
+    ITokenWithDeposit,
+    IERC677Token,
+    AccessControlled,
+    StandardToken,
+    TokenMetadata,
+    MigrationSource,
+    Reclaimable
+{
+    ////////////////////////
+    // Constants
+    ////////////////////////
+
+    string private constant NAME = "Euro Token";
+
+    string private constant SYMBOL = "EUR-T";
+
+    uint8 private constant DECIMALS = 18;
+
+    ////////////////////////
+    // Mutable state
+    ////////////////////////
+
+    // a list of addresses that are allowed to receive EUR-T
+    mapping(address => bool) private _allowedTransferTo;
+
+    // a list of of addresses that are allowed to send EUR-T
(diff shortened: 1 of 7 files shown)


contracts/EuroToken.sol
```diff
@@ -0,0 +1,235 @@
...
+            amount,
+            this,
+            extraData
+        );
+
+        return success;
+    }
+
+    //
+    // Overrides Reclaimable
```
Recmo: This whole part can be left out. It is not necessary to override if nothing is changed from the default behaviour.

contracts/EuroToken.sol
```diff
@@ -0,0 +1,235 @@
...
+
+    ////////////////////////
+    // Mutable state
+    ////////////////////////
+
+    // a list of addresses that are allowed to receive EUR-T
+    mapping(address => bool) private _allowedTransferTo;
+
+    // a list of of addresses that are allowed to send EUR-T
+    mapping(address => bool) private _allowedTransferFrom;
```
Recmo: If we want to use this contract longer, I would turn it into something more robust, like calling an external contract.

contracts/EuroToken.sol
```diff
@@ -0,0 +1,235 @@
...
+        setAllowedTransferTo(to, true);
+        Deposit(to, amount);
+        return true;
+    }
+
+    /// @notice withdraws 'amount' of EUR-T by burning required amount and providing a proof of whithdrawal
+    /// @dev proof is provided in form of log entry on which EURT_DEPOSIT_MANAGER
+    /// @dev will act off-chain to return required Euro amount to EUR-T holder
+    function withdraw(uint256 amount)
+        public
```
Recmo: We may want to put a role on here too, so we can enable/disable withdraws if we need to.

contracts/EuroToken.sol
```diff
@@ -0,0 +1,235 @@
...
+        require(_balances[msg.sender] >= amount);
+        _balances[msg.sender] -= amount;
+        _totalSupply -= amount;
+        Withdrawal(msg.sender, amount);
+    }
+
+    /// @notice enables or disables address to be receipient of EUR-T
+    function setAllowedTransferTo(address to, bool allowed)
+        public
+        only(ROLE_EURT_DEPOSIT_MANAGER)
```
Recmo: This should be a separate role, `ROLE_WHITELIST_MANAGER` or something. No reason why the entity calling `deposit` should have the rights to call this function.

contracts/EuroToken.sol
```diff
@@ -0,0 +1,235 @@
...
+    {
+        _allowedTransferTo[to] = allowed;
+        AllowedToAddress(to, allowed);
+    }
+
+    /// @notice enables or disables address to be sender of EUR-T
+    function setAllowedTransferFrom(address from, bool allowed)
+        public
+        only(ROLE_EURT_DEPOSIT_MANAGER)
```
Recmo: Same here.

contracts/EuroToken.sol
```diff
@@ -0,0 +1,235 @@
...
+    function allowedTransferFrom(address from)
+        public
+        constant
+        returns (bool)
+    {
+        return _allowedTransferFrom[from];
+    }
+
+    //
+    // Overrides ERC20 Interface to allow transfer from/to allowed addresses
```
Recmo: If we start from the `SnapshotToken`, there are internal callbacks `mOnTransfer` and `mOnApprove` that can be used. Makes it slightly simpler to implement this. (fewer functions to override, no risk that you forget one, no need to call the base class, etc.)

contracts/EuroToken.sol
```diff
@@ -0,0 +1,235 @@
...
+
+    function migrate()
+        public
+        onlyMigrationEnabled
+        onlyAllowedTransferTo(msg.sender)
+    {
+        // burn deposit
+        uint256 amount = _balances[msg.sender];
+        require( amount > 0);
+        _balances[msg.sender] -= amount;
```
Recmo: `_balances[msg.sender] = 0;`

contracts/EuroToken.sol
```diff
@@ -0,0 +1,235 @@
...
+    //
+
+    function migrate()
+        public
+        onlyMigrationEnabled
+        onlyAllowedTransferTo(msg.sender)
+    {
+        // burn deposit
+        uint256 amount = _balances[msg.sender];
+        require( amount > 0);
```
Recmo: Extra whitespace.

contracts/EuroToken.sol
rudolfix: This contract should be replaced when we have KYC and bank details verification. `_allowedTransferTo` will be replaced with a contract that holds KYC results or some other client registry. We agreed with Andre and Jacek for extremely policed version of EUR-T at start

contracts/EuroToken.sol
rudolfix: OK

contracts/EuroToken.sol
rudolfix: In current simple implementation, calling deposit whitelist `to` to receive balance. I can refactor as you wish... just in case we want to keep this contract longer.

contracts/EuroToken.sol
rudolfix: true, I will comment it out but leave the note

PR #75, opened 2017-09-07
Make commitment terms immutable state

Recmo:
Maybe we can convert some of them to constants? I'm thinking of:

* min cap
* max cap
* min ticket
* duration


rudolfix:
Totally fine with making the following proper constants: minCap, minTicket and platformOperatorWallet If we do so we can remove tests that are working with non standard values. We will not use those anyway.
I would keep duration and maxTicket as constructor parameters.
platformOperatorWallet is just simple address of our nano now and should stay this way. so can be a constant.

PR #78, opened 2017-09-14
[WIP] EuroToken commitment

Recmo:


rudolfix:
Two general remarks:
1. unordered whitelist is not implemented in Whitelisting period. This is second list from previous commitment contract. It could be emulated on ordered whitelist with 0 ticket size but currently `addWhitelistInvestorPrivate` will try to assign Neumarks for tickets with 0 amount. Is it intentional?
2. Why we have separate `commitEuro` and `commitEtherPrivate` ? they all do the same after we deposit ether into token. Why not `commitToken(investor, address, amount, amountEurUlps) private` or similar that works with any token that supports deposit?

rudolfix:
Seems that `ITokenOffering` is not used anymore in `Commitment`. In that case it does not make sense to keep it. `LockedAccount` can also be simplified (now it can be passed from one Commitment contract to another but we do not need that anymore). Should we remove it? I can handle `LockedAccount` modifications.

Recmo:
> unordered whitelist is not implemented in Whitelisting period. This is second list from previous commitment contract. It could be emulated on ordered whitelist with 0 ticket size but currently addWhitelistInvestorPrivate will try to assign Neumarks for tickets with 0 amount. Is it intentional?

Intentional. Neumark and everything else does the sensible thing when amount is zero.

Recmo:
> Why we have separate commitEuro and commitEtherPrivate ? they all do the same after we deposit ether into token. Why not commitToken(investor, token, amount, amountEurUlps) private or similar that works with any token that supports deposit?

There is commit, which everyone should use. For us there is also commitEuro.

commitEtherPrivate is there because Solidity compiler can not figure out how to use the stack otherwise.

The two functions look very similar, this is intentional. It may seem that there is a lot of overlap in code that could be factored out, but in reality, each part differs slightly because of the Eth/Eur difference. I tried merging them, but all the conditionals made it look terribly complex. Plus, we would be the only ones using the Eur part anyway.

Recmo:
> Seems that ITokenOffering is not used anymore in Commitment. In that case it does not make sense to keep it. LockedAccount can also be simplified (now it can be passed from one Commitment contract to another but we do not need that anymore). Should we remove it? I can handle LockedAccount modifications.

Correct, I haven't done this cleanup step. I also think `LogFundsInvested` can be refactored a bit. (Also, it used the word 'purchased' which I find dangerous)

I don't know what can be simplified in LockedAccount, but if you see opportunities, please do!

rudolfix:
> The two functions look very similar, this is intentional. It may seem that there is a lot of overlap in code > that could be factored out, but in reality, each part differs slightly because of the Eth/Eur difference. I tried merging them, but all the conditionals made it look terribly complex. Plus, we would be the only ones using the Eur part anyway.

I think If you would operate only on EUR values internally those functions would be identical. For example whitelist could store euroUlps only. I do not even think that we should enforce currency that is used to pay for the reserved slots. Then we can also drop token type and (native) amount from whitelist. Just keep euro values.
Also caps can be checked against euroUlps (converted wei or native).
Then code simplifies to smth like this
```
function commit()
        external
        payable
        withTimedTransitions()
        onlyStates(State.Whitelist, State.Public)
    {
        ETHER_TOKEN.transferFrom(msg.sender, this, commitedWei);

        // Turn msg.value into EtherToken (if any)
        commitedWei += msg.value;
        ETHER_TOKEN.deposit.value(msg.value)();

        // Get Neumark reward
        uint256 commitedEur = convertToEur(commitedWei);
        var (investorNmk, ticketNmk) = commitToken(commitedEur);

        // Lock EtherToken
        ETHER_TOKEN.approve(ETHER_LOCK, commitedWei);
        ETHER_LOCK.lock(msg.sender, commitedWei, investorNmk);
        _whitelistEtherNmk -= ticketNmk;

        // Log successful commitment
        LogFundsInvested(
            msg.sender,
            commitedWei,
            ETHER_TOKEN,
            commitedEur,
            investorNmk,
            NEUMARK
        );
    }

    function commitEuro()
        external
        withTimedTransitions()
        onlyStates(State.Whitelist, State.Public)
    {
        // Receive Euro tokens
        uint256 euroUlp = EURO_TOKEN.allowance(msg.sender, this);
        EURO_TOKEN.transferFrom(msg.sender, this, euroUlp);

        // Get Neumark reward
        var (investorNmk, ticketNmk) = commitToken(euroUlp);

        // Lock EuroToken
        EURO_TOKEN.approve(EURO_LOCK, euroUlp);
        EURO_LOCK.lock(msg.sender, euroUlp, investorNmk);
        _whitelistEuroNmk -= ticketNmk;

        // Log successful commitment
        LogFundsInvested(
            msg.sender,
            euroUlp,
            EURO_TOKEN,
            euroUlp,
            investorNmk,
            NEUMARK
        );
    }

    function commitToken(uint256 euroUlp)
        internal
        returns (uint256 investorNmk, uint256 ticketNmk)
    {
        // Compute commited funds
        require(euroUlp >= MIN_TICKET_EUR);
        uint256 remainingEur = euroUlp;
        uint256 totalNmk = 0;

        // Whitelist part
        WhitelistTicket storage ticket = _whitelist[msg.sender];
        bool whitelisted = ticket.amountEur > 0;
        if (whitelisted) {
            uint256 ticketEur = min(remainingEur, ticket.amountEur);
            ticketNmk = proportion(
                ticket.rewardNmk,
                ticketEur,
                ticket.amountEur);
            ticket.amountEur -= ticketEur;
            ticket.rewardNmk -= ticketNmk;
            remainingEur -= ticketEur;
            totalNmk += ticketNmk;
        }

        // Curve
        if (whitelisted || state() == State.Public) {
            totalNmk += NEUMARK.issueForEuro(remainingEur);
            remainingEur = 0;
        }

        // We don't do partial tickets
        require(remainingEur == 0);

        // We don't go over the cap
        require(NEUMARK.totalEuroUlps() <= CAP_EUR);

        // Split the Neumarks
        uint256 platformNmk = divRound(totalNmk, PLATFORM_SHARE);
        assert(platformNmk <= totalNmk);
        investorNmk = totalNmk - platformNmk;

        // Issue Neumarks and distribute
        NEUMARK.transfer(msg.sender, investorNmk);
        NEUMARK.transfer(PLATFORM_WALLET, platformNmk);

        return (investorNmk, ticketNmk);
    }
```

rudolfix:
> The two functions look very similar, this is intentional. It may seem that there is a lot of overlap in code that could be factored out, but in reality, each part differs slightly because of the Eth/Eur difference. I tried merging them, but all the conditionals made it look terribly complex. Plus, we would be the only ones using the Eur part anyway.

I think If you would operate only on EUR values internally those functions would be identical. For example whitelist could store euroUlps only. We can drop native amount from whitelist. Just keep euro values.
Also caps can be checked against euroUlps (converted wei or native).
Then code simplifies to smth like this
```
    function commit()
        external
        payable
        withTimedTransitions()
        onlyStates(State.Whitelist, State.Public)
    {
        ETHER_TOKEN.transferFrom(msg.sender, this, commitedWei);

        // Turn msg.value into EtherToken (if any)
        commitedWei += msg.value;
        ETHER_TOKEN.deposit.value(msg.value)();

        // Get Neumark reward
        uint256 commitedEur = convertToEur(commitedWei);
        var (investorNmk, ticketNmk) = commitToken(commitedEur, Token.Ether);
        _whitelistEtherNmk -= ticketNmk;

        // Lock EtherToken
        ETHER_TOKEN.approve(ETHER_LOCK, commitedWei);
        ETHER_LOCK.lock(msg.sender, commitedWei, investorNmk);

        // Log successful commitment
        LogFundsInvested(
            msg.sender,
            commitedWei,
            ETHER_TOKEN,
            commitedEur,
            investorNmk,
            NEUMARK
        );
    }

    function commitEuro()
        external
        withTimedTransitions()
        onlyStates(State.Whitelist, State.Public)
    {
        // Receive Euro tokens
        uint256 euroUlp = EURO_TOKEN.allowance(msg.sender, this);
        EURO_TOKEN.transferFrom(msg.sender, this, euroUlp);

        // Get Neumark reward
        var (investorNmk, ticketNmk) = commitToken(euroUlp, Token.Euro);
        _whitelistEuroNmk -= ticketNmk;

        // Lock EuroToken
        EURO_TOKEN.approve(EURO_LOCK, euroUlp);
        EURO_LOCK.lock(msg.sender, euroUlp, investorNmk);

        // Log successful commitment
        LogFundsInvested(
            msg.sender,
            euroUlp,
            EURO_TOKEN,
            euroUlp,
            investorNmk,
            NEUMARK
        );
    }

    function commitToken(uint256 euroUlp, Token tokenType)
        internal
        returns (uint256 investorNmk, uint256 ticketNmk)
    {
        // Compute commited funds
        require(euroUlp >= MIN_TICKET_EUR);
        uint256 remainingEur = euroUlp;
        uint256 totalNmk = 0;

        // Whitelist part
        WhitelistTicket storage ticket = _whitelist[msg.sender];
        bool whitelisted = ticket.token == tokenType;
        bool whitelistActiveForToken = tokenType == Token.Euro || state() == State.Whitelist;
        if (whitelisted && whitelistActiveForToken) {
            uint256 ticketEur = min(remainingEur, ticket.amountEur);
            ticketNmk = proportion(
                ticket.rewardNmk,
                ticketEur,
                ticket.amountEur);
            ticket.amountEur -= ticketEur;
            ticket.rewardNmk -= ticketNmk;
            remainingEur -= ticketEur;
            totalNmk += ticketNmk;
        }

        // Curve
        if (whitelisted || state() == State.Public) {
            totalNmk += NEUMARK.issueForEuro(remainingEur);
            remainingEur = 0;
        }

        // We don't do partial tickets
        require(remainingEur == 0);

        // We don't go over the cap
        require(NEUMARK.totalEuroUlps() <= CAP_EUR);

        // Split the Neumarks
        uint256 platformNmk = divRound(totalNmk, PLATFORM_SHARE);
        assert(platformNmk <= totalNmk);
        investorNmk = totalNmk - platformNmk;

        // Issue Neumarks and distribute
        NEUMARK.transfer(msg.sender, investorNmk);
        NEUMARK.transfer(PLATFORM_WALLET, platformNmk);

        return (investorNmk, ticketNmk);
    }
```

Recmo:
Looks great! When I tried this before, it had way more conditionals because of Pause and Rollback. Now it looks much better :)

Note that this drops the requirement that the whitelist needs to be filled in the right token. It does create a some minor issues with burning tokens, if you pay in a different currency than the signup, the wrong total is subtracted. If you signed up in eth, but pay in eur, it will still use your (already burned) ticket, causing some other investor to not receive his neumark.

rudolfix:
> Note that this drops the requirement that the whitelist needs to be filled in the right token. It does create a some minor issues with burning tokens, if you pay in a different currency than the signup, the wrong total is subtracted. If you signed up in eth, but pay in eur, it will still use your (already burned) ticket, causing some other investor to not receive his neumark.

look here
```
bool whitelisted = ticket.token == tokenType;
bool whitelistActiveForToken = tokenType == Token.Euro || state() == State.Whitelist;
```

this exactly prevents case you describe. forces to use declared token. ideally we would not force this but then rollback procedure gets harder

contracts/Commitment/StateMachine.sol
```diff
@@ -0,0 +1,124 @@
...
+    // Internal functions
+    ////////////////////////
+
+    function transitionTo(State newState)
+        internal
+        returns (State oldState)
+    {
+        oldState = _state;
+        if (oldState == newState) {
+            return;
```
rudolfix: shouldn't we require different state. is cycling withing state supported? if not I would do `assert(oldState != newState)` just to catch problem when inheriting contract cycles unexpectedly

contracts/Commitment/StateMachine.sol
```diff
@@ -0,0 +1,124 @@
...
+        }
+        require(validTransition(oldState, newState));
```
rudolfix: another case not to return on cycling: it may be prohibited by `validTransition` function. let it execute

contracts/Commitment/TimedStateMachine.sol
```diff
@@ -0,0 +1,101 @@
...
+
+    int256 internal constant WHITELIST_DURATION = 5 days;
+
+    int256 internal constant PUBLIC_DURATION = 30 days;
+
+    //
+    // Starting time relative to WHITELIST_START
+    //
+
+    int256 internal constant PUBLIC_START = WHITELIST_DURATION;
```
rudolfix: I would use a name that says that is it relative. Typically you expect timestamp from `XXXX_START` names

contracts/EuroToken.sol
```diff
@@ -121,8 +126,8 @@ contract EuroToken is
         public
     {
         require(_balances[msg.sender] >= amount);
```
rudolfix: then you can drop this. `sub` already checks for overflows

contracts/Commitment/ITokenOffering.sol
```diff
@@ -9,7 +9,9 @@ contract ITokenOffering {
     ////////////////////////
 
     /// on every investment transaction
-    /// `investor` invested `amount` in `paymentToken` currency which was converted to `eurEquivalent` that purchases `purchasedAmount` of `ofToken`
+    /// `investor` invested `amount` in `paymentToken` currency which was
+    /// converted to `eurEquivalent` that purchases `purchasedAmount` of
+    /// `ofToken`.
     event LogFundsInvested(
```
rudolfix: Why do we keep this interface if it is not used? To compile `LockedAccount`? It requires this interface to work.

contracts/Standards/ITokenWithDeposit.sol
```diff
@@ -1,33 +0,0 @@
-pragma solidity 0.4.15;
-
-import './IERC20Token.sol';
-
-
-contract ITokenWithDeposit is IERC20Token {
```
rudolfix: Why should we delete it? Then you duplicate definitions for log events and withdrawal. Is this because of `deposit` that is payable?

Recmo: Yes, EuroToken's deposit function is too different from EtherToken's. Different people call it, payable, etc. Also, as an interfaec, what does it mean to have a `TokenWithDeposit`? Does it mean I can now call `deposit` on it? What do I pay it with?

contracts/Commitment/ITokenOffering.sol
Recmo: Yes, it's there because it's used in LockedAccount.

contracts/Commitment/TimedStateMachine.sol
Recmo: :+1:

contracts/Commitment/StateMachine.sol
rudolfix: ok let `validTransition` handle this

Recmo: Cycling is not supported and silently ignored. I don't think turning it into a denial of service is a good idea.

contracts/Commitment/StateMachine.sol
Recmo: I prefer the current semantics of silently ignoring.

contracts/EuroToken.sol
Recmo: Sub does an `assert`, which is not appropriate here. It needs to be `required`.

contracts/Commitment/StateMachine.sol
rudolfix: heh OK, my thinking was: better reject in testing than silently push to production. it is academic discussion. this state machine is too simple to have this case even by mistake

contracts/EuroToken.sol
rudolfix: this is what Zeppelin is doing everywhere and I also do not like. But then with this check what is the point of using `sub`? We should not use math functions mindlessly...
I'm going to change Zeppelin token implementations accordingly, look
```
// Check is not needed because sub(_allowance, _value) will already throw if this condition is not met
// require (_value <= _allowance);
```

Recmo: ok.

contracts/Standards/ITokenWithDeposit.sol
rudolfix: It simply would be nice to have interface to tokens where you can deposit and withdraw assets. pity there's too many differences. I'm still for some kind of standard here. Let's merge now. I will think about it :)

contracts/EuroToken.sol
rudolfix: no worries, will be separate PR

PR #79, opened 2017-09-15
standardizes token revert behavior

rudolfix:
Standardizes EtherToken, EuroToken and Neumark behavior. All functions in ERC20, ERC677 and ERC223 that return bool can only return true. Any other case will revert.

Revert case is sending to address(0) in all tokens

With this behavior now superfluous return values in MAllowance and MMint internal interfaces were removed.

Test cases were added to ERC20, ERC677 and ERC223 and applied to all relevant tokens


contracts/test/TestERC223Callback.sol
```diff
@@ -0,0 +1,51 @@
...
+    function tokenFallback(
+        address from,
+        uint256 amount,
+        bytes data
+    )
+        public
+    {
+        _from = from;
+        _amount = amount;
+        _dataKeccak = sha3(data);
```
Recmo: It would be simpler to just fire a log event. In fact, we could have a generic mock contract that has a fallback function that logs `msg.sig` and `msg.data` and then returns the next value from a preconfigured array. That should cover a lot of cases.

PR #80, opened 2017-09-15
Check transfer results + require balance

Recmo:


contracts/Commitment/Commitment.sol
```diff
@@ -210,7 +210,7 @@ contract Commitment is
     {
         // Take with EtherToken allowance (if any)
         uint256 commitedWei = ETHER_TOKEN.allowance(msg.sender, this);
-        ETHER_TOKEN.transferFrom(msg.sender, this, commitedWei);
+        assert(ETHER_TOKEN.transferFrom(msg.sender, this, commitedWei));
```
rudolfix: why assert, nor require? this is not invariant that you check

Recmo: I'm using allowance, so it should never fail. Same for the other ones, if a transfer fails, that is a serious bug.

PR #83, opened 2017-09-19
adds agreement structure to ico

rudolfix:
splits issuing Neumarks into
1. `issueForEuro` used by NEUMARK_ISSUER to itself, no agreements are signed
2. `function distributeNeumark(address to, uint256 neumarkUlps)` used by NEUMARK_ISSUER to distribute, agreement is signed for `to` address, regardless of transferEnabled
3. Adds `Agreement` to `Commitment`

if direction is OK I will continue with the tests and deployment update

### DIFF
--- a/contracts/Neumark.sol
+++ b/contracts/Neumark.sol
@@ -70,7 +70,7 @@ contract Neumark is
         NeumarkIssuanceCurve()
         Reclaimable()
     {
-        _transferEnabled = false;
+        _transferEnabled = true;
         _totalEuroUlps = 0;
     }
 
@@ -81,6 +81,7 @@ contract Neumark is
     function issueForEuro(uint256 euroUlps)
         public
         only(ROLE_NEUMARK_ISSUER)
+        acceptAgreement(msg.sender)
         returns (uint256)
     {
         require(_totalEuroUlps + euroUlps >= _totalEuroUlps);
@@ -96,7 +97,7 @@ contract Neumark is
         only(ROLE_NEUMARK_ISSUER)
         acceptAgreement(to)
     {
-        bool success = mTransfer(msg.sender, to, neumarkUlps);
+        bool success = transfer(to, neumarkUlps);
         require(success);
     }
 
--- a/contracts/Neumark.sol
+++ b/contracts/Neumark.sol
@@ -81,7 +81,6 @@ contract Neumark is
     function issueForEuro(uint256 euroUlps)
         public
         only(ROLE_NEUMARK_ISSUER)
-        acceptAgreement(msg.sender)
         returns (uint256)
     {
         require(_totalEuroUlps + euroUlps >= _totalEuroUlps);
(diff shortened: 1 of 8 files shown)


contracts/Neumark.sol
```diff
@@ -92,6 +91,15 @@ contract Neumark is
...
 
+    function distributeNeumark(address to, uint256 neumarkUlps)
+        public
+        only(ROLE_NEUMARK_ISSUER)
+        acceptAgreement(to)
+    {
+        bool success = mTransfer(msg.sender, to, neumarkUlps);
+        require(success);
+    }
+
```
Recmo: This works, and seems secure, but I'm still not convinced by this approach.

It's behaviour is essentially that of the normal `transfer` function. The name `distributeNeumark` is not explaining why there is a second transfer function. It is also not clear why only `ISSUER` is allowed to use this function (it doesn't issue anything, so why issuer?). All of this only starts to make sense in the context of the Commitment contract. But this contract will only be live for a short while, where this function, and its odd reason d'etre, will be live for years.

The whole reason this function exists is just so that Commitment can call `acceptAgreement` on `to`as Neumark. So why not offer this function directly? We could even implement it direcetly in Agreement. It would of course have to be protected by a special role like `only(ROLE_IMPLIED_ACCEPTOR)` or something similar:

```
contract Agreement {
    …
    function agreementImplied(address acceptor)
        public
        only(ROLE_IMPLIED_ACCEPTOR)
        acceptAgreement(acceptor)
    {
    } 
```

```
contract Comitment {
    …
            // Accept agreement and transfer Neumarks
            NEUMARK.agreementImplied(msg.sender);
            NEUMARK.transfer(msg.sender, investorNmk);
            NEUMARK.transfer(PLATFORM_WALLET, platformNmk);
```

```
rbac.setUserRole(commitment, ROLE_IMPLIED_ACCEPTOR, neumark);
```

No changes to Neumark itself required. Much better future compatibility when we need to implied-accept the agreement from other contracts for other reasons that do not involve `ROLE_NEUMARK_ISSUER`.

rudolfix: What you describe above creates two additional roles and both of those breach trustless trust.
1. `ROLE_IMPLIED_ACCEPTOR` may sign any one up for any reason. This should not be at all necessary. We sign people up only due to them executing a meaningful transaction on a contract that inherits from Agreement. So current pattern in Agreement is (almost) good. In case of Neumark there are two such cases
- I claim new Neumarks that were issued to me
- I transfer Neumarks I've got from whatever or approve such transfer.
2. `TRANSFERER` role let's platform operator transfer Neumarks when other people can or (because this is very selective permission) give such rights to its buddies. This is a breach of trustless trust to me.

We are heroically fighting a wrong pattern we created :> We should force investors to claim their issued Neumarks afterer ICO. This is of course hard to do from the user experience perspective, also it is too late for the change.

I also do not agree that this problem will go away. In future ETOs we have escrow so some kind of pattern with smart contract issuing neumarks and releasing them later to investor will stay (and some function that combines transfer with sign up will be necessary)

Here what I would do now:
1. I would enable transfers by default in Neumark from the very start. Neumark will be issued and traded at the same time anyway. This will solve problems with TRANSFERER and giving permissions to enable such transfers to Commitment contract.
2. I would keep `distributeNeumarks` as is but make it to observe transfer enabled/disabled. As I said this is replacement of `function issueForEuroTo(address to, uint256 euroUlps)` that was issuing directly to investor and also signing him up (that was nice and clean but reserved slots killed it). This is also the reason only `ISSUER` may call it.

There are tons of nuances in there but overall I think we should go this way. (or quickly implement `claim` pattern. idk.)

PR #87, opened 2017-10-03
improves deployment

rudolfix:
Reorganizes deployment scripts and provides deployment documentation in README.
Fixes a typo in ACCESS_CONTROLLER role name.


yarn.lock
```diff
@@ -10,6 +10,13 @@ abbrev@1.0.x:
   version "1.0.9"
   resolved "https://registry.yarnpkg.com/abbrev/-/abbrev-1.0.9.tgz#91b4792588a7738c25f35dd6f63752a2f8776135"
 
+accepts@~1.3.4:
```
krzkaczor: Why is this file changed? there are no changes to package json. Did you updated deps or did something that would otherwise trigger this?

migrations/6_relinquish_control.js
```diff
@@ -0,0 +1,77 @@
+require("babel-register");
+const controlAccounts = require("./accounts").default;
+
+const RoleBasedAccessControl = artifacts.require("RoleBasedAccessControl");
+const EuroToken = artifacts.require("EuroToken");
+
+// Needs to match contracts/AccessControl/RoleBasedAccessControl.sol:TriState
+const TriState = { Unset: 0, Allow: 1, Deny: 2 };
```
krzkaczor: this is repeated multiple times. I would suggest extracting it to separate file and exporting it from there.

migrations/accounts.js
```diff
@@ -0,0 +1,39 @@
+// Maps roles to accounts on live network
+const ACCESS_CONTROLLER = "0xaa11C97Be40Cdf6e24229EAA731EE3701C3B9493";
+const LOCKED_ACCOUNT_ADMIN = "0xE8Bba2F6bcF6893E66F316d7f2546ee8534F2868";
+const WHITELIST_ADMIN = "0x45FE1297E62c57179BEbe2cdcBcf22b9005769b4";
+const PLATFORM_OPERATOR_WALLET = "0xfE9641B0f229a2D55a3D1aBC77eCa90648BD5357";
+const PLATFORM_OPERATOR_REPRESENTATIVE =
+  "0x11CA6fdC4117feb52d8b81adA7AdEBd519dD8D08";
+const EURT_DEPOSIT_MANAGER = "0x113b51BFAcbF6BC25B553E2a0B1C90741c25C6eB";
+
+export default function controlAccounts(network, accounts) {
```
krzkaczor: I think it would be much wiser to return object here. Then calling it and destructing it is a way more readable and self-describing.

migrations/6_relinquish_control.js
```diff
@@ -0,0 +1,77 @@
...
+
+// Maps roles to accounts on live network
+let ACCESS_CONTROLLER;
+
+module.exports = function deployContracts(deployer, network, accounts) {
+  // do not deploy testing network
+  if (network === "inprocess_test" || network === "coverage") return;
+  [ACCESS_CONTROLLER] = controlAccounts(network, accounts);
+
+  deployer.then(async () => {
```
krzkaczor: why not `await deployer`?

migrations/5_amend_agreements.js
```diff
@@ -0,0 +1,42 @@
...
+  if (network.endsWith("_live")) {
+    console.log("---------------------------------------------");
+    console.log(
+      `Must use ${PLATFORM_OPERATOR_REPRESENTATIVE} account to deploy agreements on live newtork`
+    );
+    console.log("---------------------------------------------");
+    return;
+  }
+
+  deployer.then(async () => {
```
krzkaczor: why not `await deployer`

migrations/5_amend_agreements.js
```diff
@@ -0,0 +1,42 @@
+require("babel-register");
+const controlAccounts = require("./accounts").default;
+
+const Neumark = artifacts.require("Neumark");
+const Commitment = artifacts.require("Commitment");
+
+// Maps roles to accounts on live network
+let PLATFORM_OPERATOR_REPRESENTATIVE;
+const RESERVATION_AGREEMENT =
+  "ipfs:QmPXME1oRtoT627YKaDPDQ3PwA8tdP9rWuAAweLzqSwAWT";
```
krzkaczor: I would take these hardcoded constants and the ones in accounts.js file and put them in separate file so they can be easily changed in future

yarn.lock
rudolfix: yes, this is what I did. should we rollback yarn file?

migrations/5_amend_agreements.js
rudolfix: please nooo! if you export async function from test to truffle, it will fail in so unpredictable way. we spent hours to find this pattern, which works deterministic. better not touch it now!

migrations/accounts.js
rudolfix: done!

migrations/6_relinquish_control.js
rudolfix: done!

migrations/5_amend_agreements.js
krzkaczor: sure ;D i am fine with this solution if it works reliably

yarn.lock
krzkaczor: if it works it's fine ;) I am was just curious why it changed

PR #90, opened 2017-10-15
adds legal documents

rudolfix:


legal/README.md
```diff
@@ -0,0 +1,111 @@
...
+As described in Agreement.sol, several smart contracts contain backing legal agreements. You should refer to this file for rationale and mechanism.
+You can also read README of our ESOP project here https://github.com/Neufund/ESOP#esop-terms--conditions-document
+
+Documents are available in Word (docx) and html (converted from Word, do not expect too much!).
+Documents are templates within which several tags marked with curly braces needs to be replaced before document is stored in IPFS.
+
+## Tag replacement and document lifecycle
+Tags are replaced in three stages that correspond to document lifecycle.
+1. Initial phase. Html version is taken and initial tags are replaced. **This version is stored in IPFS**
+2. General phase. Happens in context of smart contract. Html is retrieved from IPFS and general tags are replaced. Such version may be shown as general terms, to be read before user makes the transaction. (you can think of nice replacements of personal tags, see below).
```
krzkaczor: Why we split initial phase and general phase? General phase variables seem like something that would be set right after all contracts are deployed.

rudolfix: Initial phase documents will be stored into IPFS so those tags are replaced "forever". General phase you do into commit page. Is explanation not clear?

PR #91, opened 2017-10-15
adds user signature block number

rudolfix:
Instead of simple bool, block numbers that carry agreement signing dates are stored in Agreement
tests updated accordingly

### DIFF
--- a/contracts/Agreement.sol
+++ b/contracts/Agreement.sol
@@ -50,8 +50,11 @@ contract Agreement is
     // Mutable state
     ////////////////////////
 
+    // stores all amendments to the agreement, first amendment is the original
     SignedAgreement[] private _amendments;
-    mapping(address => bool) private _signatories;
+
+    // stores block numbers of all addresses that signed the agreement (signatory => block number)
+    mapping(address => uint256) private _signatories;
 
     ////////////////////////
     // Events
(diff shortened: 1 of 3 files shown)


contracts/Agreement.sol
```diff
@@ -51,7 +51,7 @@ contract Agreement is
     ////////////////////////
 
     SignedAgreement[] private _amendments;
-    mapping(address => bool) private _signatories;
+    mapping(address => uint256) private _signatories;
```
krzkaczor: Maybe leave a comment here that it's address -> block number mapping? or rename it to idk signatoriesToBlockNumber?

test/Agreement.js
```diff
@@ -66,9 +66,13 @@ contract(
       expect(event.args.accepter).to.eq(signer);
     }
 
-    async function expectAgreementAccepted(signer) {
-      const signed = await agreement.isAgreementSignedBy.call(signer);
-      expect(signed).to.be.true;
+    async function expectAgreementAccepted(signer, firstTx) {
```
krzkaczor: why is this parameter named `firstTx`? I think name like `signTx` makes more sense

contracts/Agreement.sol
rudolfix: so true

PR #93, opened 2017-10-18
Audit of Commitment subsystem

chfast:
The audit was done from 3-10-2017 to 20-10-2017, 55 hours of work in total.

The following contract were reviewed:
- MStateMachine,
- StateMachine,
- TimedStateMachine,
- Commitment.

This list of commit contains comments, suggestions and issues in chronological order.

No critical issues found.

### DIFF
--- a/contracts/Commitment/Commitment.sol
+++ b/contracts/Commitment/Commitment.sol
@@ -460,6 +460,18 @@ contract Commitment is
     {
         // Validate
         require(investor != 0x0);
+
+        // AUDIT[CHF-34] Undocumented enum feature used.
+        //   In the following line you are relaying on the undocumented enum
+        //   feature that the first element of an enum is also the "zero"
+        //   element and storage variables of enum types are initialized to
+        //   "zero" elements.
+        //   This is correct: confirmed by solidity developers and community
+        //   (https://ethereum.stackexchange.com/questions/21775/what-is-the-zero-value-for-an-enum).
+        //   Moreover, considering EVM internals there is no
+        //   practical way for solidity to implement this differently.
+        //   A contribution to solidity documentation describing this behavior
+        //   would be a nice addition to this ICO project.
         require(_whitelist[investor].token == Token.None);
         bool isEuro = token == Token.Euro;
         bool isEther = token == Token.Ether;
--- a/contracts/Commitment/Commitment.sol
+++ b/contracts/Commitment/Commitment.sol
@@ -587,6 +587,10 @@ contract Commitment is
         // Split the Neumarks
         uint256 platformNmk = divRound(totalNmk, PLATFORM_SHARE);
         assert(platformNmk <= totalNmk);
+
+        // AUDIT[CHF-49] Safe Math.sub() unnecessary below.
+        //   It is guaranteed that totalNmk >= platformNmk.
+        //   The same issue is in whitelistTicket().
         investorNmk = sub(totalNmk, platformNmk);
 
         // Issue Neumarks and distribute
--- a/contracts/Commitment/Commitment.sol
+++ b/contracts/Commitment/Commitment.sol
@@ -111,6 +111,13 @@ contract Commitment is
     event LogFundsCommited(
         address indexed investor,
         uint256 amount,
+
+        // AUDIT[CHF-28] Group indexed arguments.
+        //   On the Ethereum RPC level this value is going to be
+        //   passed as a LOG topic together with the first indexed argument
+        //   `investor`. This might be confusing that the order of values in RPC
+        //   is different that the oder of arguments in this declaration.
+        //   Consider putting indexed arguments before non-indexed ones.
         address indexed paymentToken,
         uint256 eurEquivalent,
         uint256 grantedAmount,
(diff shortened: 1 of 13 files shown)


contracts/Commitment/Commitment.sol
```diff
@@ -210,22 +262,57 @@ contract Commitment is
...
+        //   accepting the agreement. Shouldn't the order of actions be
+        //   different? Accepting the agreement being a precondition of commit()
+        //   action?
         acceptAgreement(msg.sender)
     {
         // Take with EtherToken allowance (if any)
         uint256 commitedWei = ETHER_TOKEN.allowance(msg.sender, this);
+
+        // AUDIT[CHF-54] Unnecessary call to ETHER_TOKEN.transferFrom().
+        //   When commitedWei is 0, calling ETHER_TOKEN.transferFrom() will
```
rudolfix: There is minimum ticket size which is 1 ETH now, so in the case you describe contract will revert. Surely, gas will be used but that's going to be user's fault. Blockchain will not be polluted with unnecessary log entries.

It may be good idea to move minimum ticket checks before transfers and deposits are made: move it from `commitToken` and duplicat in `commit` and `commitEuro`. That would make contract clear.

contracts/Commitment/Commitment.sol
```diff
@@ -181,23 +217,39 @@ contract Commitment is
...
 
             // Loop body is factored out to keep stack low
+            // AUDIT[CHF-31] Invalid assumption about EVM stack space.
+            //   This pattern will not save you any EVM stack space (probably
+            //   the opposite). Solidity does not use CALLs to execute private
+            //   functions. It will just JUMP to the referenced function
+            //   staying on the same call depth -- so the stack space is
+            //   shared.
+            //   To use a separated call for execution of this function you have
+            //   to use `this.addWhitelistInvestorPrivate()`, but
```
rudolfix: this also costs 700 gas units for `CALL`

contracts/Commitment/Commitment.sol
```diff
@@ -276,6 +363,17 @@ contract Commitment is
     {
         uint256 amountEur = convertToEur(amountEth);
         uint256 rewardNmk = NEUMARK.incremental(amountEur);
+        // AUDIT[CHF-47] Investor's share calculation inconsistency.
```
rudolfix: Great catch!

contracts/Commitment/Commitment.sol
```diff
@@ -210,22 +262,57 @@ contract Commitment is
...
+        // AUDIT[CHF-58] Move NMK counters updates to Commitment.commitToken().
+        //   The Commitment.commitToken() has all the logic related to whitelist
+        //   checking. Move the _whitelistEtherNmk subtraction from here and
+        //   _whitelistEuroNmk subtraction from commitEuro() there to
+        //   commitToken() too. This will also simplify the return type of
+        //   commitToken().
         _whitelistEtherNmk = sub(_whitelistEtherNmk, ticketNmk);
 
         // Lock EtherToken
+        // AUDIT[CHF-60] Consider using ERC223-like transfer pattern.
```
rudolfix: We had long debate over it here. Personally I do not like ERC223 because you lose explicit semantics in the contract receiving tokens. It's ok to use it on simple token transfers where `transfer` is the main operation of the contract. Here the function is to lock funds. `ETHER_TOKEN.transfer(...)` does not say anything what is being done in the lock. We'd need to encode `investorNmk` in the data field as well.

contracts/Commitment/Commitment.sol
```diff
@@ -210,22 +262,57 @@ contract Commitment is
...
+        //   The onTokenTransfer() callback can be implemented in the
+        //   LockedAccount contract to lock tokens received by ERC223-like
+        //   transfers. Still we will have 2 calls to lock tokens, but
```
rudolfix: 2 calls? I'm probably not getting this.

contracts/Commitment/Commitment.sol
```diff
@@ -362,6 +469,7 @@ contract Commitment is
         return (ticket.token, ticket.amountEur, investorNmk);
     }
 
+    // AUDIT[CHF-46] What is this function for?
```
rudolfix: to enumerate whitelist

contracts/Commitment/Commitment.sol
```diff
@@ -413,6 +527,18 @@ contract Commitment is
     {
         // Validate
         require(investor != 0x0);
+
+        // AUDIT[CHF-34] Undocumented enum feature used.
```
rudolfix: That is true. I'd rather have explicit semantics that map enums to values. This is also the reason we have strict requirement on `solc` version for compilation and testing.

contracts/Commitment/Commitment.sol
```diff
@@ -464,24 +615,72 @@ contract Commitment is
...
+        //      AUDIT[CHF-63].
+        //
+        //   The proposed changes are in audit/CHF-64.patch.
         require(remainingEur == 0);
 
         // We don't go over the cap
         require(NEUMARK.totalEuroUlps() <= CAP_EUR);
 
         // Split the Neumarks
+        // AUDIT[CHF-52] Simplify calculating the platform share.
```
rudolfix: OK, we need to also update rounding scheme in the tests

contracts/Math.sol
```diff
@@ -1,6 +1,6 @@
 pragma solidity 0.4.15;
 
-
+// AUDIT[CHF-37] The Math contract should have a dedicated unit test suite.
```
rudolfix: Yeah! I'll port it from our ESOP project.

PR #95, opened 2017-10-20
Add nano admin-tool

Moejoe90:


truffle.js
```diff
@@ -2,6 +2,39 @@ require("babel-register");
 require("babel-polyfill");
 const TestRPC = require("ethereumjs-testrpc");
 
+/**
+  In order to deploy normally without this setup you will have to comment all the code from =====>
+*/
+const Web3 = require("web3");
```
rudolfix: can you put this intialization subroutine into /scripts as a function and then import it using `require`?

truffle.js
```diff
@@ -2,6 +2,39 @@ require("babel-register");
...
+
+engine.addProvider(new FilterSubprovider());
+engine.addProvider(
+  LedgerWalletSubproviderFactory.default(new Web3(web3HttpProvider), nanoPath)
+);
+engine.addProvider(new Web3Subprovider(web3HttpProvider));
+engine.start();
+/**
+  =====> Till here
+  BUG: truffle cannot finish deployment without commenting the above section
```
rudolfix: this is because of maybe too short timeout on Commitment contract START date... it is not important with automatic deployment

truffle.js
```diff
@@ -46,7 +79,8 @@ module.exports = {
     simulated_live: {
```
rudolfix: can you make additional network called `nano` and leave this intact?

truffle.js
```diff
@@ -46,7 +79,8 @@ module.exports = {
...
       network_id: "*",
       host: "localhost",
-      port: 8545
+      port: 8545,
+      provider: engine // Our costume instance
```
rudolfix: 1. `custom` not `costume`
2. when you import nano support as above just do `provider: nano.start()` or smth similar

truffle.js
Moejoe90: on that

truffle.js
```diff
@@ -1,7 +1,16 @@
+/* eslint-disable global-require */
 require("babel-register");
 require("babel-polyfill");
 const TestRPC = require("ethereumjs-testrpc");
 
+const providerUrl = "http://localhost:8545";
+const nanoPath = "44'/60'/0'/0`";
+
+const nanoProvider = () =>
```
rudolfix: can you make providerUrl and nanoPath parameters to nanoProvider? and then instantiate properly later? then we can have many instances with different parameters

truffle.js
```diff
@@ -58,8 +67,8 @@ module.exports = {
       network_id: "*",
       host: "localhost",
```
rudolfix: not necessary anymore - passed in custom provider

truffle.js
```diff
@@ -58,8 +67,8 @@ module.exports = {
...
       port: 8545,
```
rudolfix: not necessary anymore - passed in custom provider

truffle.js
```diff
@@ -1,7 +1,16 @@
+/* eslint-disable global-require */
 require("babel-register");
 require("babel-polyfill");
 const TestRPC = require("ethereumjs-testrpc");
 
+const providerUrl = "http://localhost:8545";
+const nanoPath = "44'/60'/0'/0`";
+
+const nanoProvider = () =>
+  process.argv.filter(arg => arg.includes("nano")).length > 0
```
rudolfix: add another parameter to nanoProvider = (...., networkName) and filter exact arg => arg === networkName so we can instantiate just one reacting to network name that we are requesting

scripts/AccessControl.js
```diff
@@ -0,0 +1,14 @@
+/* eslint-disable no-console */
```
rudolfix: please remove this file from this pull request. it is a different task

truffle.js
Moejoe90: I had an idea similar to _live we can create _nano and this should check if _nano was included this.
Very similar to what is happening in `config.js`

truffle.js
Moejoe90: done

scripts/AccessControl.js
Moejoe90: done

truffle.js
Moejoe90: done

truffle.js
Moejoe90: You mean from the arguments that will be a tricky approach hence the idea mentioned below `_nano` we can set some flags and check for the parameter after the flag we will just need to make sure it doesn't activate truffle somehow.

rudolfix: smth like this
```
const nanoProvider = (providerUrl, derivationPath, network) =>
  process.argv.filter(arg => arg === network).length > 0
    ? require("./nanoWeb3Provider").nanoWeb3Provider(providerUrl, derivationPath)
    : undefined;
```

rudolfix: then we can create many instances like
`provider: nanoProvider("http://localhost:8545",  "44'/60'/0'/0", "nano")`

truffle.js
```diff
@@ -2,6 +2,14 @@ require("babel-register");
 require("babel-polyfill");
 const TestRPC = require("ethereumjs-testrpc");
 
+const providerUrl = "http://localhost:8545";
+const nanoPath = "44'/60'/0'/0`";
```
rudolfix: what is this "`" doing here ;>

truffle.js
```diff
@@ -65,10 +68,8 @@ module.exports = {
     },
     nano: {
       network_id: "*",
-      host: "localhost",
-      port: 8545,
       gas: 4600000,
-      provider: nanoProvider()
+      provider: nanoProvider("http://localhost:8545", "44'/60'/0'/0`", "nano")
```
rudolfix: remove this little ` sign from derivation path

truffle.js
```diff
@@ -3,13 +3,16 @@ require("babel-register");
...
 
-const providerUrl = "http://localhost:8545";
-const nanoPath = "44'/60'/0'/0`";
-
-const nanoProvider = () =>
-  process.argv.filter(arg => arg.includes("nano")).length > 0
-    ? require("./nanoWeb3Provider").nanoWeb3Provider(providerUrl, nanoPath)
-    : undefined;
+const nanoProvider = (providerUrl, nanoPath, network) => {
+  for (let i = 0; i < process.argv.length; i += 1) {
```
rudolfix: why don't you use nice list operations like `some`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some then it is one liner

Moejoe90: Looks better indeed

PR #97, opened 2017-10-22
commitment audit processed

rudolfix:


contracts/Commitment/Commitment.sol
```diff
@@ -612,11 +546,7 @@ contract Commitment is
             ticket.rewardNmk = sub(ticket.rewardNmk, ticketNmk);
             remainingEur = sub(remainingEur, ticketEur);
 
-            // AUDIT[CHF-62] Usage of "+=" instead of "=".
-            //  The totalNmk is always 0 in this place.
-            //  Use `totalNmk = tickerNmk;` or use totalNmk variable instead of
-            //  ticketNmk in this code block.
-            totalNmk = add(totalNmk, ticketNmk);
+            totalNmk += ticketNmk;
```
chfast: You sill can use `totalNmk = ticketNmk` here. I need confirmation you want to keep the `+=`.

contracts/Commitment/Commitment.sol
```diff
@@ -349,6 +340,7 @@ contract Commitment is
         constant
         returns (uint256)
     {
+        require(amount < 2**123);
```
chfast: Where is the number `2**123` coming from?

contracts/Commitment/Commitment.sol
```diff
@@ -563,6 +556,13 @@ contract Commitment is
         NEUMARK.distributeNeumark(msg.sender, investorNmk);
         NEUMARK.distributeNeumark(PLATFORM_WALLET, platformNmk);
 
-        return (investorNmk, ticketNmk);
+        if (ticketNmk > 0) {
```
chfast: Maybe it's better to move this logic inside `if (whitelisted && whitelistActiveForToken) {` block?

contracts/Commitment/Commitment.sol
```diff
@@ -565,4 +540,15 @@ contract Commitment is
         }
         return investorNmk;
     }
+
+    // calculates investor's and platform operator's neumarks from total reward
+    function calculateNeumarkDistribtion(uint256 rewardNmk)
+        private
+        returns (uint256 platformNmk, uint256 investorNmk)
+    {
+        // round down - platform may get 1 wei less than investor
```
chfast: Do you have a test case where it actually happens?

contracts/Commitment/StateMachine.sol
```diff
@@ -96,4 +105,9 @@ contract StateMachine is MStateMachine {
             oldState == State.Public && newState == State.Finished
         );
     }
+
+    /// @notice gets called after every state transition.
+    /// @dev may not change state, transitionTo will revert on that condition
+    function mAfterTransition(State oldState, State newState)
```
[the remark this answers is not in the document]
rudolfix: yes, I decided to keep it. there is no additional risk posed by having this parameter and it generalizes interface nicely.

contracts/Commitment/Commitment.sol
rudolfix: There is explanations two lines above. I'll added a small details there, now as follows
```
/// @notice Considering the max possible ETH_EUR_FRACTION value (10**18*10**4 == ~2**73), the max
    ///       amount of ETH (not wei) that is safe to be passed as the argument
    ///       is ~10**(54 - 18) (~2**123).
```

contracts/Commitment/Commitment.sol
```diff
@@ -548,32 +548,26 @@ contract Commitment is
...
-        //   The general rules are described in
-        //   "Order of operations within an external or public function"
-        //   in CodeStyle.md file.
-        //
-        //   At least `_whitelist[investor].token = token` should be set before
-        //   calling NEUMARK.issueForEuro().
-        //
-        //   The proposed code changes are in audit/CHF-42.patch.
-        //
-        //   Also, having a unit test case for reentracy attack on
```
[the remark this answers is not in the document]
rudolfix: While I totally support the pattern to modify state and then call external contracts, in that case re-entrancy is not possible, even if Neumark contract calls back into this or another function, the `msg.sender` will be different and it will be kicked out by `only(ROLE_WHITELIST_ADMIN)` of `addWhitelisted`. Similar situation with other calls to Neumark.

contracts/Commitment/Commitment.sol
rudolfix: added in #5c7fbb1

contracts/Commitment/Commitment.sol
rudolfix: nice simplification! in 74cd932

contracts/Commitment/Commitment.sol
```diff
@@ -493,17 +450,13 @@ contract Commitment is
...
-            //   you will also get some gas back.
-            //   The same applies to _whitelistEuroNmk when
-            //   newState == State.Finished.
+            _whitelistEtherNmk = 0;
         }
         if (newState == State.Finished) {
 
             // Rollback unfulfilled Euro reservations.
             NEUMARK.burnNeumark(_whitelistEuroNmk);
+            _whitelistEuroNmk = 0;
```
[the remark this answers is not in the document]
rudolfix: also simplifies code per 9152285

contracts/Commitment/Commitment.sol
```diff
@@ -425,6 +426,8 @@ contract Commitment is
             ETHER_LOCK.controllerSucceeded();
             EURO_LOCK.controllerSucceeded();
         }
+        // burn Neumarks after state change to prevent theoretical re-entry
+        NEUMARK.burn(nmkToBurn);
```
chfast: As in the other cases, you probably don't want to trigger the burning if there is nothing to burn (`nmkToBurn == 0`).

rudolfix: in this case I'd like to have a trace that there was attempt to burn 0 Ulps of Neumark. This means that whole whitelist was realized by investors. This is actually useful information. I've checked. Logs will be generated.

chfast: Ok, but you will get a similar log also for transition to Whitelist state.

contracts/Commitment/Commitment.sol
```diff
@@ -332,64 +355,69 @@ contract Commitment is
...
 
     //
-    // MStateMachine
+    // Implements StateMachine
     //
 
     function mAfterTransition(State /* oldState */, State newState)
         internal
     {
+        uint256 nmkToBurn;
```
chfast: I noticed that usually you explicitly initialize local variables.

PR #98, opened 2017-10-22
Legal docs improvements

krzkaczor:
- readme changes
- script to easily fill documents
- **add missing getter** SOLIDITY CHANGE
- update ipfs hashes to point to updated documents (better for testing etc)

### DIFF
--- a/contracts/Commitment/Commitment.sol
+++ b/contracts/Commitment/Commitment.sol
@@ -342,6 +342,14 @@ contract Commitment is
         return MIN_TICKET_EUR;
     }
 
+    function ethEurFraction()
+        public
+        constant
+        returns (uint256)
+    {
+        return ETH_EUR_FRACTION;
+    }
+
     function platformOperatorNeumarkRewardShare()
         public
         constant
--- a/legal/README.md
+++ b/legal/README.md
@@ -44,7 +44,6 @@ For personal phase, block number at which user signed agreement may be obtained 
 |{commit-id}|Commit-id for which deployed byte code (referenced by this agreement) may be reproduced|
 |{website}|Commitment website|https://commit.neufund.org|
 |{acquisition-sc-address}|Commitment contract address|*input parameter*|
-|{signed-by-company-date}|Timestamp (UTC) at which agreement was signed by company|Commitment::currentAgreement[1]|
 |{company-address}|Ethereum address that signed agreement|Commitment::currentAgreement[0]|
 |{neumark-sc-address}|Neumark contract address|Commitment::neumark|
 |{icbm-start-date}|When public ICBM starts|Commitment::startOf(State.Public) (timestamp)|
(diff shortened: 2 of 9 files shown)


contracts/Commitment/Commitment.sol
```diff
@@ -342,6 +342,14 @@ contract Commitment is
         return MIN_TICKET_EUR;
     }
 
+    function ethEurFraction()
```
rudolfix: Why just not call `convertToEur` with  1 ether? If you persist on having this (there may be numeric precision problems due to mul/div) please add test for getter here:
```
it("should deploy", async () => {
      await prettyPrintGasCost("Neumark deploy", neumark);
--> here
    });
```
in `Neumark.js`
later I'll merge audit results with other getters' tests

legal/README.md
```diff
@@ -42,34 +43,35 @@ For personal phase, block number at which user signed agreement may be obtained
...
-|{fork-arbiter-sc-address}|Address of fork arbiter contract|Commitment::ethereumForkArbiter|
+|{reservation-period}|Lock period in seconds|LockedAccount::lockPeriod|
 
 **Personal phase**
 
 |Tag|Description|Value|
 |---|-----------|-----|
 |{investor-address}|Address of investor for which Agreement is generated|*input parameter*|
-|{payment-token}|Token which investor uses to commit|LockedAccount::assetToken()::name|
+|{current-block-hash}|Hash of current block|*input parameter*|
```
rudolfix: More precise description
Hash of the block from which all other values were taken. 
Please pay attention to take all values from that block really (not just use "latest" which will just take current block) to avoid inconsistency.

legal/formatters.js
```diff
@@ -0,0 +1,41 @@
+// eslint-disable-next-line
+const BigNumber = require("bignumber.js");
```
rudolfix: is this duplicated in `commitment` page repo?

legal/prepareDocuments.js
```diff
@@ -0,0 +1,93 @@
+require("babel-register");
```
rudolfix: is this duplicated in `commitment` page repo?

krzkaczor: No - this file is not duplicated in commitment at all.

legal/formatters.js
krzkaczor: this particular no — commitment is in typescript. These functions are also in commitment page but generally, there are more of them.

legal/prepareDocuments.js
```diff
@@ -0,0 +1,93 @@
...
+  neumarkContract,
+  companyAddress
+}) {
+  const neuDecimals = (await neumarkContract.decimals()).toNumber();
+
+  const tags = {
+    "repo-url": DOCUMENT_CONSTANTS.repoUrl,
+    "commit-id": gitRev.long(),
+    "neumark-sc-address": neumarkContract.address,
+    "signed-by-company-date": formatMomentDate(moment.utc()),
```
rudolfix: this should get this value from contract as per documentation:  `Neumark::currentAgreement[1]`

legal/prepareDocuments.js
```diff
@@ -0,0 +1,93 @@
...
+  lockedAccountContract,
+  companyAddress
+}) {
+  const tags = {
+    "repo-url": DOCUMENT_CONSTANTS.repoUrl,
+    "commit-id": gitRev.long(),
+    website: DOCUMENT_CONSTANTS.website,
+    "acquisition-sc-address": commitmentContract.address,
+    "lockin-sc-address": lockedAccountContract.address,
+    "signed-by-company-date": formatMomentDate(moment.utc()),
```
rudolfix: this should get this value from contract as per documentation:  `Commitment::currentAgreement[1]`

PR #99, opened 2017-10-23
fixes role based access policy permissions initialization

rudolfix:


contracts/AccessControl/RoleBasedAccessPolicy.sol
```diff
@@ -108,11 +111,11 @@ contract RoleBasedAccessPolicy is
         revert();
     }
 
-    // Implements `IAccessPolicy.allowed(IAccessControlled, bytes32, address, bytes4)`
+    // Implements `IAccessPolicy.allowed(address, bytes32, address, bytes4)`
```
krzkaczor: why you're replacing IAccessControlled with address? Isn't implementing that interface required for a contract to work with our RBAC?

rudolfix: No it is not, it is just prerequisite for using `only` modifier that indeed works on contract itself. but nothing prohibits assigning permissions to contracts that do not need `IAccessControlled::allowed()`

PR #104, opened 2017-10-26
Audit of LockedAccount

chfast:


contracts/LockedAccount.sol
```diff
@@ -302,6 +335,8 @@ contract LockedAccount is
         return PENALTY_FRACTION;
     }
 
+    // AUDIT[CHF-132] Rename LockedAccount.balanceOf().
```
rudolfix: investor balance here is a tuple... what's wrong with it?

contracts/SnapshotToken/Helpers/Allowance.sol
```diff
@@ -80,6 +80,9 @@ contract Allowance is
         public
         returns (bool success)
     {
+        // AUDIT[CHF-116] Unnecessary require() in Allowance.approveAndCall()
```
rudolfix: This is why we dropped all `false` return value forwarding/bubbling up in all of our tokens. Whenever there is `false` we immediately revert. Already in master branch... 
However I didn't think about that case. Good catch. We base our implementation on MiniMe token. I'll let them know about the problem.

contracts/LockedAccount.sol
chfast: Not a big deal, but one of the values is the unlock date. It does not match the "balance" name so much.

chfast:
I agree with other comments.

contracts/LockedAccount.sol
```diff
@@ -175,21 +183,43 @@ contract LockedAccount is
...
+            //   in each lock() call?
             a.unlockDate = currentTime() + LOCK_PERIOD;
         }
+
+        // AUDIT[CHF-108] Unnecessary storage update in LockedAccount.lock().
+        //   The following line is not needed because `a` is already the
+        //   reference to the storage location.
+        //   Please observer gas cost changes. I noticed that the cost of
+        //   Commitment.commit() increased after removing the following line.
         _accounts[investor] = a;
```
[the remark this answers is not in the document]
rudolfix: Checked with Remix
1. with optimization enabled there is just one SSTORE no matter if `_accounts[investor] = a;` is there or not. When line is deleted the gas cost increases as you observed. All that increase comes from execution, not from storage access, didn't dive into optimized assembly too deeply
2. with optimization disable there are 4 SSTORE instructions vs 1 SSTORE when line is removed. so all as expected!

I'm removing this anyway

PR #117, opened 2017-11-03
Moe/ipfs uploader

Moejoe90:
This PR passes with all required specification
- More work needs to be done with cleaning code in terms File reading and error handling
- Only tested with private local node but should work 
`yarn uploadAgreements "ipfs.neustg.net" "./legal/NEUMARK TOKEN HOLDER AGREEMENT.out.html"`
or default
`yarn uploadAgreements`

### DIFF
--- a/scripts/uploadAgreements.js
+++ b/scripts/uploadAgreements.js
@@ -34,7 +34,8 @@ const addFiletoIpfs = async (ipfs, file) => {
   }
   return false;
 };
-
+// TODO: handle diffrent ports for ipfs
+// TODO: clean code
 const main = async args => {
   const ipfsNodeAddress = args[0];
   const filePaths = [];
--- a/scripts/uploadAgreements.js
+++ b/scripts/uploadAgreements.js
@@ -1,0 +1,15 @@
+/* eslint-disable no-console */
+
+// eslint-disable-next-line import/no-extraneous-dependencies
+require("babel-register");
+const ipfsAPI = require("ipfs-api");
+
+// eslint-disable-next-line no-unused-vars
+module.exports = async function(callback) {
+  const ipfs = ipfsAPI("ipfs.neustg.net:5001");
+  try {
+    console.log(await ipfs.pin.ls());
+  } catch (err) {
+    console.log(err);
+  }
+};
--- a/scripts/uploadAgreements.js
+++ b/scripts/uploadAgreements.js
@@ -1,15 +1,72 @@
 /* eslint-disable no-console */
 
 // eslint-disable-next-line import/no-extraneous-dependencies
-require("babel-register");
 const ipfsAPI = require("ipfs-api");
+const fs = require("fs");
+const path = require("path");
+const Promise = require("bluebird");
 
-// eslint-disable-next-line no-unused-vars
-module.exports = async function(callback) {
-  const ipfs = ipfsAPI("ipfs.neustg.net:5001");
+const isFilePinned = async (ipfs, hash) => {
+  const matchedHashes = (await ipfs.pin.ls()).filter(response => {
+    if (response.hash === hash) return true;
+    return false;
+  });
+  return matchedHashes.length > 0;
+};
+const addFiletoIpfs = async (ipfs, file) => {
   try {
-    console.log(await ipfs.pin.ls());
+    let fileHash = (await ipfs.files.add(file, { "only-hash": true }))[0].hash;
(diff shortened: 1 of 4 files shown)


scripts/uploadAgreements.js
```diff
@@ -0,0 +1,85 @@
+/* eslint-disable no-console */
+
+// eslint-disable-next-line import/no-extraneous-dependencies
+const ipfsAPI = require("ipfs-api");
+const fs = require("fs");
+const path = require("path");
+
+const isFilePinned = async (ipfs, hash) => {
+  const matchedHashes = (await ipfs.pin.ls()).filter(response => {
```
rudolfix:
```
{
 +    if (response.hash === hash) return true;
 +    return false;
 +  }
```
may be replaced with single `response.hash === hash`

scripts/uploadAgreements.js
```diff
@@ -0,0 +1,85 @@
...
+        file: fs.readFileSync(filePath)
+      };
+    } catch (e) {
+      throw new Error(`Can't read file "${filePath}"`);
+    }
+  });
+
+// TODO: handle diffrent ports for ipfs
+const main = async ([ipfsNodeAddress, ...paths]) => {
+  const defaultDirpath = path.join(__dirname, "..", "legal");
```
rudolfix: rename to `defaultDirPath`

scripts/uploadAgreements.js
```diff
@@ -0,0 +1,85 @@
...
+  const filePaths = [];
```
rudolfix: fileNames

scripts/uploadAgreements.js
```diff
@@ -0,0 +1,85 @@
...
+  const defaultfilePaths = [
```
rudolfix: `defaultFileNames` paths are not names. also mind capitalization

scripts/uploadAgreements.js
```diff
@@ -0,0 +1,85 @@
...
+    "NEUMARK TOKEN HOLDER AGREEMENT.out.html",
+    "RESERVATION AGREEMENT.out.html"
+  ];
+  try {
+    if (!paths) throw new Error("Please give ipfs node");
```
rudolfix: huh? you should check `ipfsNodeAddress` for being empty here?

scripts/uploadAgreements.js
```diff
@@ -0,0 +1,85 @@
...
+        name: path.parse(filePath).base,
+        file: fs.readFileSync(filePath)
+      };
+    } catch (e) {
+      throw new Error(`Can't read file "${filePath}"`);
+    }
+  });
+
+// TODO: handle diffrent ports for ipfs
+const main = async ([ipfsNodeAddress, ...paths]) => {
```
rudolfix: are those `paths` or `fileNames`?

scripts/uploadAgreements.js
```diff
@@ -0,0 +1,85 @@
...
+const isFilePinned = async (ipfs, hash) => {
+  const matchedHashes = (await ipfs.pin.ls()).filter(response => {
+    if (response.hash === hash) return true;
+    return false;
+  });
+  return matchedHashes.length > 0;
+};
+
+const addFiletoIpfs = async (ipfs, file, name) => {
+  try {
```
rudolfix: do we need try...catch here? just let exception go. you will catch it later

scripts/uploadAgreements.js
Moejoe90: good point i missed this one when refactoring code

scripts/uploadAgreements.js
Moejoe90: Done

scripts/uploadAgreements.js
Moejoe90: done

scripts/uploadAgreements.js
Moejoe90: Done

scripts/uploadAgreements.js
Moejoe90: These are indeed paths

scripts/uploadAgreements.js
Moejoe90: Done

scripts/uploadAgreements.js
Moejoe90: This is actually file paths

PR #125, opened 2017-11-11
whitelisting script

Moejoe90:


scripts/uploadWhitelist.js
```diff
@@ -0,0 +1,177 @@
...
+};
+const isCurrency = (currency, address) => {
+  if (currency !== "EUR" && currency !== "ETH")
+    throw new Error(
+      `Investor:${address} and currency:${currency} has wrong currency format!!`
+    );
+  return tokenEnum[currency];
+};
+const getAmount = (amount, address) => {
+  const investAmount = Number.parseFloat(amount.replace(/^\D+/g, ""));
```
rudolfix: we need to use better parser. this one accepts any messy string starting with a number

scripts/uploadWhitelist.js
```diff
@@ -0,0 +1,177 @@
...
+
+const getAttributes = (address, currency, amount) => ({
+  wlAddresses: isAddress(address),
+  wlTokens: isCurrency(currency, address),
+  wlTickets: getAmount(amount, address)
+});
+const isDuplicate = array => {
+  const duplicateArray = array.filter(
+    data =>
+      array.filter(investor => {
```
rudolfix: use `some`, `undefined` should not be returned

scripts/uploadWhitelist.js
```diff
@@ -0,0 +1,177 @@
...
+  }
+};
+const filterWlfromSmartContract = async (filteredWhiteList, commitment) => {
+  let lastUploadedfile = false;
+  const verifiedWhiteList = (await Promise.all(
+    filteredWhiteList.map(async investor => {
+      const whiteListTicketbyAddress = await commitment.whitelistTicket(
+        investor.wlAddresses
+      );
+      const tokenType = whiteListTicketbyAddress[0].isZero();
```
rudolfix: what is `isZero()`?

scripts/uploadWhitelist.js
```diff
@@ -0,0 +1,177 @@
...
+      const ticketSize = whiteListTicketbyAddress[1].toString();
+      const ticketSizefromList =
+        investor.wlTokens === 1
+          ? (await commitment.convertToEur(investor.wlTickets)).toString()
+          : investor.wlTickets.toString();
+
+      if (ticketSize !== ticketSizefromList && tokenType === false) {
```
rudolfix: use bigNumber.eq() not !==

rudolfix: `tokenType === false` what does it mean here. if tokenType === 0 we should immediately show that expected investor is not on the list

scripts/uploadWhitelist.js
```diff
@@ -0,0 +1,177 @@
...
+    })
+  )).filter(investor => {
+    if (investor !== undefined) {
+      lastUploadedfile = true;
+    }
+    if (investor === undefined) {
+      if (lastUploadedfile === true)
+        throw new Error("Something went really wrong with the smart contracts");
+    }
+    return !!investor;
```
rudolfix: what !!means in javascript? two negations?

scripts/uploadWhitelist.js
```diff
@@ -0,0 +1,177 @@
...
+      commitment
+    );
+    do {
+      const whitelistPayloud = verifiedWhiteList.slice(0, payloadSize);
+      verifiedWhiteList = verifiedWhiteList.slice(
+        payloadSize,
+        verifiedWhiteList.length
+      );
+
+      console.log("Setting whitelist Payloud");
```
rudolfix: Payload

scripts/uploadWhitelist.js
```diff
@@ -0,0 +1,177 @@
...
+      const formattedPayload = formatPayload(whitelistPayloud);
+      console.log("Sending Payload to SmartContract");
+      const transactionReceipt = (await commitment.addWhitelisted(
+        formattedPayload[0].wlAddresses,
+        formattedPayload[1].wlTokens,
+        formattedPayload[2].wlTickets
+      )).receipt;
+      if ((await transactionReceipt.status) === 0)
```
rudolfix: always show receipt

scripts/uploadWhitelist.js
```diff
@@ -0,0 +1,177 @@
...
+            ticketSizefromList
+          } token ${tokenType} for ${investor.wlAddresses}`
+        );
+      }
+      return tokenType ? investor : undefined;
+    })
+  )).filter(investor => {
+    if (investor !== undefined) {
+      lastUploadedfile = true;
+    }
```
rudolfix: why not else here?

scripts/uploadWhitelist.js
```diff
@@ -0,0 +1,177 @@
...
+    if (investor === undefined) {
```
rudolfix: what this one means? if we added an investor and now we see that someone is already added, we throw an error?

Moejoe90: Of course this makes sure that the list is in order. 
If we find someone already added before there que an error is thrown

scripts/uploadWhitelist.js
Moejoe90: Its automatically done by the linter if not negetive return true

